PR #563, opened 2014-07-12
Update contribute.rmd

andrewpbray:
I'm new to contributing via GitHub and I was a bit timid as the directions didn't quite mesh with what I was seeing (no obvious way to 'save', the description can be written before the pull request).  It now coheres with my process of contribution (hopefully generalizable).

A period was added to the end of the first item for consistency.


hadley:
Thanks!

PR #588, opened 2014-08-12
Some unexpected behavior I tested in R

LaDilettante:
The behavior described about NA index is not correct. It is not a straightforward case of disallowing NA + integer indices but allowing NA + logical indices. Please see examples.


hadley:
Thanks. I think this needs a bit more explanation so I've made an issue to come back to later.

PR #590, opened 2014-08-15
Update Functions.rmd

zross:
Tiny, but isn't the colon an infix operator? I don't think I see single colon here.


hadley:
Thanks - made the change by hand since it was easier than resolving the merge conflict.

PR #592, opened 2014-08-19
Fixing a typo

grettke:
Changed `factor` to `factory`


hadley:
Looks like it got fixed in the meantime - thanks for your contribution :smile:

grettke:
😄

PR #609, opened 2014-10-02
Update Subsetting.rmd

matthieugomez:


hadley:
Thanks. I don't think it's good practice to use `which()` for its side effect of removing `NA`, but I'll add that when I come back to this section.

PR #635, opened 2014-12-14
Update Functions.rmd

mbrmbr:
Just return "a" instead of printing to be consistent with other examples, and consistent with following paragraph, where it says "returns the same value".


hadley:
Thanks!

PR #638, opened 2014-12-31
Update Functionals.rmd

davechilders:


davechilders:
My modification removes trailing NA values (which is the behavior I would expect). Here's a small example:

x <- 1:10

rollmean(x, 2) # rollmean() as defined in your book
##  [1]  NA 1.5 2.5 3.5 4.5 5.5 6.5 7.5  NA  NA

rollmean2(x, 2) # modified function
##  [1]  NA 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5

hadley:
Oh hmmm, I think I was thinking that the NAs should be symmetrical (which I obviously failed at), but now that I think about it more, your solution seems correct.

hadley:
Thanks!

Issue #644, opened 2015-01-09
UI: The left panel bleeds onto the main content on wide screens


daattali:
I'm not sure if this is an inherit problem in bootstrap JS or if it's specific to this book-site.

When I render the site on a wide monitor (currently 1920px), the left-hand panel that serves for navigation + announcements is initially ok.  But after scrolling down (on any page), that column gets fixed to the top of the browser, and its width changes. It happens on all screen resolutions, but I never noticed it because only on wide screens it would change enough to actually go on top of the content.  It's not too terrible, but it does make it a bit less readable so I just wanted to bring it to your attention.

I took a screenshot of how it looks at the top of the page vs scrolling down a bit:
![navbug](https://cloud.githubusercontent.com/assets/952340/5673519/cfcd8450-9751-11e4-91f3-1fe0f738e87c.png)

<br/>

__The cause__  
The navigation column is set to have width 25%.  When we're at the top of the page, that means 25% of its parent container.  But when scrolling down, the navigation gets detached from its parent with `position: fixed`, which means that now the 25% is referring to 25% of the viewport, which is much larger.

To confirm that this is indeed what's happening, here's a snippet from the JS console, starting when the page is at the top

![navconsole](https://cloud.githubusercontent.com/assets/952340/5673443/e18ee432-9750-11e4-8639-98b6bcd61e9c.png)

hadley:
Ah, thanks for the debugging help - I had noticed that the navbar width wasn't constant but I hadn't figured out why. I notice that http://getbootstrap.com/javascript/#affix uses a pixel width for their navbar, so I should probably do the same

hadley:
That fix works for me locally, but I don't have a very wide screen. Can you please test in ~15 minutes? (once travis has rebuilt the site)

daattali:
Yup that solution works fine.  
The only thing (non-major) is that this ignores the responsiveness of bootstrap and will use the same width no matter how large the content area is.  I might suggest a slightly improved css later, but thanks for addressing

PR #645, opened 2015-01-09
Make navigation width responsive to screen width

daattali:
https://github.com/hadley/adv-r/issues/644


hadley:
Thanks!

hadley:
Would you mind doing the same for r-pacakges?

PR #647, opened 2015-01-11
small bugfixes in beyond-exception-handling.rmd

Mullefa:
* from `malformed_log_entry_error()` I inferred that `...` in `condition()` should be used to pass further elements to the underlying `list` which is returned

* some anonymous functions used in `tryCatch()`, `withRestarts()` and `withCallingHandlers()` didn't take any arguments, whereas in fact they should take one


Mullefa:
Sorry, I realise what's happening now.

hadley:
Thanks!

PR #650, opened 2015-01-14
Update dsl.rmd

JoeVoelkel:
In your escape notes (lines 123-130), it seems that the results need to look like the following. I'm guessing right now that displaying this in html is the source of the problem... (I entered "jgv" in the file itself at this point, so you can see it and so I can click Propose file change.)

> escape("x > 1 & y < 2")
<HTML> x &gt; 1 &amp; y &lt; 2
> escape(escape("This is some text. 1 > 2"))
<HTML> This is some text. 1 &gt; 2
> escape(html("<hr />"))
<HTML> <hr />

### DIFF
--- a/dsl.rmd
+++ b/dsl.rmd
@@ -118,7 +118,7 @@ escape.character <- function(x) {
 escape.list <- function(x) {
   lapply(x, escape)
 }
-
+jgv
 # Now we check that it works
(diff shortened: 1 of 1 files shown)


hadley:
This looks like a bug introduced with a recent version of knitr. Thanks for pointing it out!

PR #652, opened 2015-01-16
Update Performance.rmd

JoeVoelkel:
"validated", yes?
four, yes? I see 32 to 8 ms in the benchmarking


hadley:
It's invalidated - the stored data needs to be marked as no longer valid.

I'll look into the performance numbers when I'm working on that section.

PR #670, opened 2015-04-22
Variable Name Change

pankajpansari:
Changing variable name from 'c' to 'd' in order to avoid confusion with R's built in c() command.


hadley:
In this context, I don't think it should be confusing.

PR #671, opened 2015-04-25
names() can return NA

gezakiss7:
Described the case when names() returns NA, as a result of adding some, but not all names after the creation of a vector.


hadley:
Hmmmm, this seems like a buglet in R

hadley:
Thanks! This inconsistency definitely needs to be described somewhere.

PR #677, opened 2015-05-05
method name <-> class name

gezakiss7:
In `t.test`, if it was an S3 method, "test" would refer to the class, not "t'.


hadley:
Thanks!

PR #680, opened 2015-05-08
Install Rexamine/stringi from GitHub

wibeasley:
https://github.com/Rexamine/stringi/issues/155


hadley:
I think it would be better to fix the root cause.

PR #682, opened 2015-05-23
Made source of `subs()` clear.

iangow:
The function `pry::subs()` is used before the fact that it comes from `pryr` is made clear.


iangow:
Reading the hard copy of the book, I realized that I was somehow reading out of order (clicking a "back arrow" in a browser can give confusing results). So ignore this one.

PR #684, opened 2015-05-29
Update Profiling.rmd

carey1024:
the argument to Sys.sleep in
system.time(parLapply(cluster, 1:10, function(i) Sys.sleep(1)))
should be i, not 1


hadley:
Thanks!

PR #685, opened 2015-06-04
Update OO-essentials.rmd

DamienFoucard:
I suggest to replace all abbreviated forms of the verb be, i.e., " 's ", by the corresponding full form, namely " is ". I believe it would make this documentation more professional and slightly clearer.


hadley:
I prefer the more conversation style with abbreviations.

DamienFoucard:
Alright, why not? Thanks for the answer anyway.
I wish you a nice week-end.

Best,
Damien

hadley:
Eliminating contractions feels needless formal to me

PR #695, opened 2015-08-05
Update Function-operators.rmd

yportne:
download_file should be the argument that pass through the function operators.


hadley:
Thanks - I modified by hand since I wanted a slightly different style

PR #696, opened 2015-08-06
more specific

isomorphisms:
Which part of the documentation?

Maybe even mention `width.cutoff`…


hadley:
Thanks! `width.cutoff` is the second argument and is pretty suspicious sounding, so I don't think it needs another hint

PR #700, opened 2015-09-01
Changed example to check if log(x) is possible

FvD:
It seems to make more sense to check whether x < 0 before doing a log(x) than so check whether x == 0 as this will yield no result other than `-Inf`.


hadley:
It was supposed to allude to the box-cox transformation.

PR #705, opened 2015-10-07
custom_stop, not customStop

mbojan:
A typo in the section on custom conditions. Accidentally it produced the correct result in the output.


hadley:
Oops, thanks!

Issue #672, opened 2015-04-28
Lazy evaluation example no longer works


csgillespie:
The code (http://adv-r.had.co.nz/Functions.html lazy evaluation)

``` r
> add <- function(x) {
+   function(y) x + y
+ }
```

is used to illustrate the issues with lazy evaluation. However, running the code, doesn't reproduce the output in the book

``` r
R> adders <- lapply(1:10, add)
R> adders[[1]](10)
[1] 11
R> adders[[10]](10)
[1] 20
```

``` r
R> sessionInfo()
R version 3.2.0 (2015-04-16)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 14.04.2 LTS

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8    
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8    LC_PAPER=en_GB.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] colorout_1.1-0

loaded via a namespace (and not attached):
[1] tools_3.2.0    fortunes_1.5-2
```

hadley:
Yes, this is a change in R 3.2

csgillespie:
I guessed that. I suppose I should have asked, do you want "error" reports that are due to R/package  changes.

jefferys:
This confused me too, until I found an associated [stack-exchage comment](http://stackoverflow.com/questions/16129902/explain-a-lazy-evaluation-quirk).

hadley:
Duplicate of #740

PR #707, opened 2015-10-21
Not the same results from R and the book

gumeo:
I added a space in the empty line 552 to be able to propose this "change" or otherwise report this bug/issue:

When I run the following code in R:

add <- function(x) {
  function(y) x + y
}
adders <- lapply(1:10, add)
adders[[1]](10)
adders[[10]](10)

I get:

> adders[[1]](10)
[1] 20
> adders[[10]](10)
[1] 20

But the output in the book online is:

adders[[1]](10)
#> [1] 11
adders[[10]](10)
#> [1] 20

I assume the 11 should be a 20 to demonstrate the difference of using force.

### DIFF
--- a/Functions.rmd
+++ b/Functions.rmd
@@ -549,7 +549,7 @@ f(stop("This is an error!"))
 ```
 
 This is important when creating closures with `lapply()` or a loop:
-
+ 
 ```{r}
 add <- function(x) {
   function(y) x + y


gumeo:
I am not sure if this is the appropriate etiquette of reporting something like this.

hadley:
Filing an issue would be more appropriate.  (And there's already #672)

gumeo:
Thank you! I am benefiting greatly from your book, so I guess this is probably my only change of saying thanks for your work :)

PR #709, opened 2015-10-25
I think this should be 99

xiaofeiyangyang:
Hi,Mr Wickham 
I read your book, it's very useful for me, and some skill is very cool.
when I read  "Functional programming" chapter, I think in your book it's should be 99 in you example.


hadley:
That's there deliberately to illustrate the problem with copying and pasting.

xiaofeiyangyang:
Oh I see now! thank you!

PR #710, opened 2015-10-28
Fixes #697, the garbled exercise involving plyr::.

### LINKED ISSUE #697: Computing on the Language: Question 4.4
Hi,

I'm working through the book, and I'm not sure about the intent of this question. I believe that the function in `plyr` is missing. Maybe it's a typo? Is `plyr`'s `quote` function actually the intended reference?

> What’s the escape hatch for `ggplot2::aes()`? What about `plyr::()`? What do they have in common? What are the advantages and disadvantages of their differences?

jennybc:
Apparently I had some other small edits lying around; I consult this book often!


hadley:
Thanks!

PR #716, opened 2015-12-08
Use getMethod to avoid S4 method dispatch

agrueneberg:
Hello Hadley,

I think in this case it should be `getMethod` instead of `findMethod` to avoid S4 method dispatch. `findMethod` returns the environment (as a list) that contains the method and not the method itself, and I don't know how to get the method from the environment. Please clarify.

Cheers,
Alex


hadley:
Oops, yes, thanks!

PR #718, opened 2015-12-17
Changed travis.yml to prevent build failures

sebastian-c:
At the moment, all travis builds are failing because the latest versions of the jekyll and mime-types require a version of ruby newer than that on the server. I've specified versions of these packages that depend on the old version or ruby.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - pandoc --version
 
   # Install jekyll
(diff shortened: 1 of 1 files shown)


sebastian-c:
`mime-types` has a later version here: https://rubygems.org/gems/mime-types/versions/2.99

hadley:
All fixed now. Thanks for the patch :)

PR #720, opened 2015-12-31
Profiling.Rmd: Updating typos to s5q6

michaelquinn32:
The function should have both the data frame and number of observations passed as arguments. Currently, it has neither.

Same as #706, except this request maintains `q` as an argument and removes the global dependency on `n`.

### DIFF
--- a/Profiling.rmd
+++ b/Profiling.rmd
@@ -539,7 +539,7 @@ microbenchmark(
     df <- data.frame(a = rnorm(n), b = rnorm(n))
 
     cor_df <- function(q, n) {
-      i <- sample(seq(n), n * 0.01)
+      i <- sample(seq(n), n * 0.01, replace = TRUE)
       cor(q[i, , drop = FALSE])[2,1]
     }
     ```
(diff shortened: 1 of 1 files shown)


Profiling.rmd
```diff
@@ -538,7 +538,7 @@ microbenchmark(
     n <- 1e6
     df <- data.frame(a = rnorm(n), b = rnorm(n))
 
-    cor_df <- function(i) {
+    cor_df <- function(q, n) {
       i <- sample(seq(n), n * 0.01)
```
hadley: Shouldn't this be `i <- sample(seq(n), replace = TRUE)`?

michaelquinn32:
You're right, and thanks for pointing that out. It wouldn't be bootstrapping without sampling with replacement. The most recent commit fixes that too.

hadley:
Do you need the `n * 0.01`? I have no idea what I was thinking when I wrote that.

michaelquinn32:
Sorry for the huge delay, but your comment brought up something really interesting that I wanted to research better. [If you care, that's here](http://michaelquinn32.github.io/bootstrap/). Otherwise, yeah, standard bootstrapping says sample the whole vector. You don't actually have to go that way, but it requires some corrections. 

I think this leads to a longer conversation that doesn't do much to enhance the textbook, so it's better left for another place. I updated the code to fall in line with standard bootstrapping.

hadley:
Thanks!

PR #721, opened 2016-01-04
fix typo in package name

renkun-ken:
`plyr::partial()` should be `pryr::partial()`.


hadley:
Thanks!

pro100mir:
Mute the thread

PR #722, opened 2016-01-15
Added default & non default behavior for subsetting.

xsankar:


hadley:
Not sure what you're getting at here. Maybe file an issue?

PR #726, opened 2016-01-23
Changed form of "additional"

rweyant:
I could be mistaken about the form used here, but I think this sentence makes more sense using "addition" rather than "additional."


hadley:
Ooops, thanks!

PR #727, opened 2016-01-25
Update Functionals.rmd

johnthomas12:
I changed line 845 from levels to labels. Previously, the code changed the mtcars$am vector from 0's and 1's to all <NA>. Now it appropriately changes it to auto/manual.


hadley:
Oops, thanks!

PR #730, opened 2016-01-31
christmas trees coming down in february

ajdamico:


hadley:
Thanks - fixed before I spotted your PR.

PR #732, opened 2016-02-11
Clarify restrictions when signalling conditions

krlmlr:
The "this" in the original text confused me, I have made the text more verbose.

The example now also obeys the suggested "best practice".


hadley:
Thanks - this all is likely to get substantially revised for the 2nd ed.

PR #733, opened 2016-02-12
Append effect of which() that ignores NAs in the logical vector

mokjpn:
- Fix a typo in line 599 (columns)
- which() ignores NAs in the logical vector, while ‘[‘ does not ignore
  it. I appended a sentence of this difference.


hadley:
Thanks - someone else did a PR for the same change. FWIW I think it's a bad idea to use `which()` for its side effect of dropping missing values.

PR #734, opened 2016-02-13
Differences between environments and lists

ijlyttle:
This set of proposed changes is admittedly pedantic - but I noticed that you made the point earlier that the objects live outside the environment. This is a very useful point.

I tried to find a way to rephrase the first two items such that they could not confuse this earlier point.


hadley:
That's great - thanks!  Advanced R is all about being pedantic 😁

PR #741, opened 2016-03-30
Update Functions.rmd

emjonaitis:
Currently the lazy evaluation examples using adders and adders2 appear in your text to give identical output (which they don't when I run the code on my own; there, I get the expected wrong answers for adders). My best guess is that it's because you used the name add() for both function definitions, so I propose changing the second function name. I can't explain why the later definition gets applied in earlier code, though (which makes me feel like I'm failing a pop quiz on this material).

In case it is relevant: I assign the copyright of this contribution to Hadley Wickham.


hadley:
The problem is that R has changed behaviour since this was written. I'll fix shortly

PR #742, opened 2016-04-03
Clarify the effects of unspecified vector names

paolieri:
Three methods are presented for setting vector names, but empty strings will be added for missing names only using c(). If names() or setNames() are used, the unspecified names will be NA.


hadley:
Thanks - I merged a PR with similar language.

PR #743, opened 2016-04-05
Update Functional Programming.Rmd

salayatana66:
lines 708 added a -1
line 712 changed length(coef)-1 to ifelse(open, n, n-1)

Reason: the way the function is written it never evaluates at the last point b. I tested on newton_cotes(7,32,12,32,7)


hadley:
Re-reading the wikipedia page, I'm pretty sure the current code is correct.

PR #745, opened 2016-04-22
Wrong output (6 instead of 7)

irudnyts:
If the code snippet is not evaluated, than x remains the same, as in 789 line.
In other words, the second element remains 6, because line 799 is not evaluated.
Therefore, printing x in line 814 will return "10 6 3 ..." instead of assumed "10 7 3 ..."


hadley:
Thanks!

PR #750, opened 2016-06-08
replace logical T and F with TRUE and FALSE in Subsetting.rmd

jooolia:
Replaced one character versions of TRUE and FALSE for clarity and for reasons alluded too in Style.rmd line 60 (i.e. that T and F are not reserved words).


hadley:
Thanks!

PR #752, opened 2016-06-20
Update Rcpp.rmd: obsolete C functions

ktpazder:
The absolute value function, abs(), listed in the "Other useful functions" section caused errors.  After some searching, I discovered this post: https://github.com/RcppCore/Rcpp/issues/167, identifying abs() as the old C function and std:abs() as its replacement.  I'm unsure if this is the only function that has changed.


hadley:
Thanks - I need to look into in more detail so I've created an issue.

PR #755, opened 2016-07-28
Modify example based on changes to apply.

jolars:
The apply() functions no longer (https://stat.ethz.ch/pipermail/r-announce/2015/000583.html) does lazy evaluation in the way that the text suggests, leading two of the code snippets to produce the same output. However, the unintended consequences of lazy evaluation still apply to for loops, so I modified the examples to consider for loops instead of lapply.


hadley:

PR #756, opened 2016-08-08
Add 2 (missing?) ways to access a single value

chambm:
The existing text says it compares seven ways, but there are only five ways. I was interested in performance of .subset and $+[[, so I added these. I was surprised that $carb[[32]] was no faster than $carb[32]. A very interesting case!


hadley:
I think that was just a mistake in the text.  I wouldn't expect `x[[1]]` to be any faster than `x[1]` when x is an atomic vctor

PR #760, opened 2016-09-09
an error in chapter Functionals

kfeng123:
The code of vapply2 has the problem of not only `out` should be transposed but also the initiation of out is wrong in essense.


hadley:
Ooops, thanks!

PR #761, opened 2016-09-09
Remove parens from attribute names

agrueneberg:
The text is talking about attribute names, not names of accessor functions.


hadley:
Thanks!

PR #766, opened 2016-10-04
Wrong output in code chunk due to eval = FALSE

irudnyts:
If eval = FALSE, then in the next code chunk the second element of the vector `x` will be 6, instead of 7. It must be 7, because we have changed this value by `x[2] <- 7L`.


hadley:
Fixed in another PR - thanks!

PR #767, opened 2016-10-04
Incomplete example of lazy evaluation

irudnyts:
In the example the function call returns 11, which is supposed to be 20 by the text of the book.
Probably, this is a recent change in R in the lazy evaluation.


hadley:

PR #769, opened 2016-10-06
Incomplete example of lazy evaluation

irudnyts:
The function returns 5.5 instead of "expected" 55. Probably, lapply use a black magic to force the argument by itself.


hadley:

PR #771, opened 2016-10-07
Sugesstion to use infix functions

irudnyts:
I guess to use infix functions will make it more readable and natural. (And instead of not one can use Negate to stay coherent).


hadley:
Thanks for the idea, and I agree that it improves readability, but I'd rather focus on one idea at a time.

PR #773, opened 2016-10-16
`length` of a matrix is not the same as its number of rows.

giuscri:
```
> length(m) == nrow(m) * ncol(m)
[1] TRUE
> length(m) == nrow(m) || length(m) == ncol(m)
[1] FALSE
```


hadley:
I wasn't trying to imply that they're equal, more that instead of a single function, you now have two.

PR #775, opened 2016-10-25
lapply is NOT equivalent to for-loop

inuritdino:
Lines 551-578. Somehow the `lapply` function is not equivalent to the `for` loop in this case. You know better, but the two outputs of the list of closures `adders` and `adders2` is the SAME even in your book.
Perhaps, the reason is that `lapply` calls .Internal (maybe not, you know better), but there is something fundamentally different between simple `for`-loop and `lapply` in this case. Therefore, the lazy evaluation is to be demonstrated only with the `for`-loops and _not_ `lapply` function.
Thanks!

### DIFF
--- a/Functions.rmd
+++ b/Functions.rmd
@@ -548,13 +548,16 @@ f <- function(x) {
 f(stop("This is an error!"))
 ```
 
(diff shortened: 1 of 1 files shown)


hadley:

PR #779, opened 2016-11-29
change outdated ?Classes to ?setClass

helmingstay:
The latter now contains detailed discussion of virtual classes


hadley:
Thanks!

PR #789, opened 2017-01-23
Change markdown code for the in_dir function

SolomonRoberts:
Function was not saving the current working directory as a variable for the on.exit function to reset the directory back to.


hadley:
`setwd()` invisibly returns the current wd

PR #793, opened 2017-01-28
Typo fix

Deleetdk:
Typo in function name. Could not find a local copy because my latex does not work.


Deleetdk:
There is a line more:

`(download_file() is a simple wrapper around utils::download.file() which provides a reasonable default for the file name.)`

However, I cannot find any such function and no package has been loaded in that chapter yet. I notice now that there are many other instances of it which also need to be renamed.

hadley:
You seem to have made the changes in the latex version rather than the source .Rmd.

(Also `download_file()` is deliberately defined in Function Operators)

PR #807, opened 2017-03-05
edit intro to S3.Rmd

lindbrook:


hadley:
Thanks! It's still pretty rough down lower, so might be too early for a detailed read

PR #808, opened 2017-03-27
Subassignments when RHS length is multiple of LHS

DAXaholic:
It is also possible to perform a subassignment when the length of the
LHS items is a multiple of the RHS items.


hadley:
Thanks — I ended up using some wording about recycling since that's what's happening here.

PR #823, opened 2017-07-06
point out typos: 'howver' and 'subtclass'

paleolimbot:
not a glorious edit but hopefully a useful one. fantastic book!


hadley:
Thanks!

PR #824, opened 2017-07-06
fix typo in 2nd paragraph

ChrisMuir:
Minor typo fix in the second paragraph of file `S3.Rmd`. This sentence:
"Since S3 has few built-in constraints, the key its successful use is apply the constraints yourself."
is now:
"Since S3 has few built-in constraints, the key to its successful use is applying the constraints yourself."


hadley:
Thanks!  (In the future, you don't need to describe the change - I can see it in the PR)

PR #828, opened 2017-07-06
Mention list/scalar possibility for classes early.

jimvine:
Seems like it would be handy to mention this at first occurrence.


hadley:
Good idea, thanks!

PR #830, opened 2017-07-06
S3 typos and FIXMEs

mtmorgan:
Minor typos and FIXMEs, plus a little commentary on ... in methods /generics. I assign the copyright of this contribution to Hadley Wickham


S3.Rmd
```diff
@@ -31,17 +31,18 @@ print(unclass(f))
...
+A __generic__ and its __methods__ are functions that operate on classes. The role of a  generic is to find the right method for the for the arguments that it is provided; this is sometimes called __dispatch__.  A method is a function that implements the generic behaviour for a specific class. S3 methods are function with a special naming scheme, `generic.class()`. For example, the Date method for the `mean()` generic is called `mean.Date()`, and the factor method for `print()` is called `print.factor()`. This is the reason that most modern style guides discourage the use of `.` in function names: it makes them look like S3 methods. For example, is `t.test()` the `t` method for `test` objects?
 
-You can find some S3 methods (those in the base package and those that you've created) by typing their names. However, this will but will not work with most packages because S3 methods are not exported: they live only inside the package, and not available from the global environment. Instead, you can use `getS3method()`, which will work regardless of where the method lives:
+<!-- FIXME: it seems like getS3method() is old-school here -->
```
hadley: Is there a better base R function? Or do you mean I should have a wrapper in S3?

S3.Rmd
```diff
@@ -461,6 +463,7 @@ foo.character <- function(x, ..., truncate = 10) {
 foo("This is a long string", trncate = 2)
 ```
 
+<!-- FIXME: but checking that ... is empty defeats the 'rationale' for ... in the first place -- to allow arguments to be passed to next methods, in 'your' method or some other method on the same generic -->
```
hadley: Oh shoot, you're right. I forgot about that problem.

I'm still fairly convinced that you `...` should be the default, otherwise you fundamentally limit the extensibility of the methods. But silently eating misnamed arguments is a big problem. I don't think `NextMethod(...)` will work because it modifies the existing call.

But I definitely need a more nuanced discussion here

S3.Rmd
mtmorgan: I honestly can't remember what I was thinking, sorry for the noise.

hadley: Np, thanks for all the other fixes!

PR #831, opened 2017-07-09
Rephrasings, clarifications, and corrections

seaaan:

I hope you don't mind that I was liberal in suggesting changes. My philosophy is that a proposed revision is useless only if it's not made, so I'd rather suggest a change and have you decide to ignore it than not suggest a change that you might have found useful.

I will make some line comments to point out a few places where I felt confused. One general comment is that some topics, in particular generic functions, methods, and dispatch are introduced in several places. These topics can be confusing, so it's good to repeat them. However, it also felt a bit fragmented to me. 

Thanks for encouraging feedback, for sharing the content for free (though I did buy the print copy of the 1st ed!), and for all you do for the R community.

### DIFF
--- a/S3.Rmd
+++ b/S3.Rmd
@@ -13,7 +13,7 @@ library(S3)
 
 ## Basics {#s3-basics}
 
-An S3 object is built on top of a base type with the "class" attribute set. The base type is typically a vector, though we will see later that it is also possible to create scalar-style classes, where the base types are lists. For example, take the factor. It is built on top of an integer vector, and the value of the class attribute is "factor". It stores information about the "levels" in another attribute.
+An S3 object is built on top of a base type with the "class" attribute set. The base type is typically a vector (though we will see later that it is also possible to create other types of classes). For example, take the factor. It is built on top of an integer vector, and the value of the class attribute is "factor". It stores information about the "levels" in another attribute.
 
 ```{r}
 f <- factor("a")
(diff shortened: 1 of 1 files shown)


S3.Rmd
```diff
@@ -54,7 +54,7 @@ getS3method("mean", "Date")
 ### Exercises
 
 1.  The most important S3 objects in base R are factors, data frames,
-    and date/times (Dates, POSIXct, POSIXlt). You've already seen the 
+    and date-times (Dates, POSIXct, POSIXlt). You've already seen the 
     attributes and base type that factors are built on. What base types and 
     attributes are the others built on?
```
seaaan: I think the Venn diagram (Q3) might be beneficial for you to include in the book as a figure, rather than an exercise, as it's confusing to keep them straight and a visual might help.

seaaan:
It's not easy to add line comments on lines where I didn't make a change, so I'll include my comments here. I include a brief quote from your text to identify the line I'm talking about and then my comment is inset below. I think I put them all in order. 

"I recommend using lower case class names to avoid this problem"
> Maybe don't use Date as the example for the first constructor. It has a couple distractions from the main point: the weird capitalization-naming issue and the extra cognitive effort to think about the representation as integers and the constructor taking an integer rather than a string.

"In base R, the most important example of this style is lm..."
> It's a bit confusing to be learning that scalar-style classes use named lists instead of attributes, but then to see `attributes(mod)` to investigate the structure of the first example class. Perhaps `names(mod)`, `class(mod)`

"...we get a warning when the object is printed" 
> No warning is printed

"generic.foo"
> I found this example to be too abstract. I needed some more hand-holding in terms of text describing what I was supposed to get out of it. I also felt that the `generic.skeleton issue` was a distraction and that given that you previously made the point about not using `.` in names, it would be better to avoid it by choosing a different name.

"...this seems mildly dangerous to me"
> Why? And how does double to factor coercion relate to implicit/explicit coercion? I guess this section wasn't explicit enough for me :)

"every as_x generic should have..."
> Why?

"we'll discuss default methods more in the next section"
> I suggest revisiting the `as_factor` example there and moving down the default method to there, so as not to confuse readers who haven't heard of default methods yet. 

"It does this by creating a vector of function names, `paste0("generic", ".", c(class(x), "default"))`, and looking for each method in turn." 
> This description I think might be too much of an implementation detail, could be clearer just to say that it looks for methods named `generic.class` for each class in `class(x)`, then `generic.default`. Especially since you haven't gotten to inheritance section where it's explained that `class(x)` can be more than a single string, so that might surprise readers. For the following POSIX example, explain where POSIXt comes from. 

"The methods are different, and R calls the internal method with a warning. \n One method is internal, in which case R calls the other method."
> I didn't really follow the difference between these two possibilities. 

Hope these comments are helpful!

hadley:
Thanks for the comments - they were very helpful! I merged them manually since the there was no longer a clean merge, and some of your tweaks goaded me to provide even better wording. 

I've fixed a few of your other comments and will work on the others.

seaaan:
Great, I'm glad to hear that it was helpful!

PR #834, opened 2017-07-11
Warn that function names containing dot "." may confuse package's NAMESPACE

maurolepore:


hadley:
I think this is out of scope here (because adv-r doesn't cover information about packages), and generally I don't think it is a good idea to build the namespace automatically.

PR #835, opened 2017-07-13
typo: that that

maurolepore:


hadley:
Thanks for pointing this out - I decided it needed a slightly different fix.

PR #840, opened 2017-07-26
S4: Fix typo

alandipert:
Behold, another typo. Along the way I noticed both "initialise" and "initialize" and wanted to suggest going with "initialize" as that's the name of the generic function.


hadley:
Yeah, that's probably a good idea.

PR #849, opened 2017-08-03
FIXMEs and more substantial suggestions

mtmorgan:


hadley:
BTW if you don't create a new branch, all the commits go into the original PR too.

S4.Rmd
```diff
@@ -190,6 +217,8 @@ Note that this example uses another form of `new()`: unnnamed arguments are take
 
 S4 allows you to define a "validity" method that checks if your object is valid. You set this method by calling `setValidity()`. Here's a port of our factor validity checking to S4. A validity method should return a character vector containing a list of all of the problems. If there are no problems, it should return a character vector of length 0.
 
+<!-- FIXME: generally, one would like to discourage direct slot access, providing accessors (e.g., plain-old-functions) instead, and using these throughout, including in validateFactor -->
```
hadley: It's not ok to use direct slot access in your own methods?

mtmorgan: you're probably your own most frequent user, at least initially and when the implementation is most fluid, so benefit the most from separating interface from implementation.

S4.Rmd
```diff
@@ -52,6 +52,21 @@ Slots and contains can specify the names of S4 classes, S3 classes (if registere
 
 To create a class, you call `setClass()`, supplying these three properties. Lets make this concrete with an example. Here we create two classes: a person with character `name` and numeric `age`, and an `Employee` that inherits slots and methods from `Person`, adding an additional `boss` slot that must be a `Person`.
 
+<!-- FIXME: the underlying model here is of two tables 'Person' and 'Employee' in a database. The 'isa' relationship is an implicit join. You've modelled rows, which means that each instance of the class is a scalar. But R works well on vectors. Effective S4 would model columns, and each instance would contain vectors -- 'People' and 'Employees', though since R uses singular integer(), etc, then 'Person' and 'Employee' but with the understanding that these are of length 0, 1, ...
```
hadley: Hmmm, I need to think about this more.  If I do continue down the vector route, what do I need to to make `Person` behave as a data frame column?

But I might steer away from this example altogether because a person is an object in the real world, and so reference semantics seem more natural.

S4.Rmd
```diff
@@ -143,6 +158,16 @@ As a user, you should generally not extract values out of someone else's object
 
 We can override the default behaviour by providing an `initialize()` method. You'll learn more about the details of creating methods in the next section, but in brief, the following code defines a `Person` method for the `initialize` generic.
 
+<!-- FIXME: initialize has a complicated contract that is not easily satisfied; most experienced Bioconductor developers don't actually write initialize methods, but rather a constructor Person() that checks / etc and then calls new() (like Factor(), below). For instance, S4Vectors defines 35 classes but no initialize methods, IRanges has 85 classes and no initialize methods.
```
hadley: So what are subclasses supposed to do? How do you reuse the code in the parent constructor efficiently?

S4.Rmd
hadley: Can you point to any antecedents for this in other programming languages?

mtmorgan: Suppose you change from a literal in-memory representation with slots `name`, `age` to an on-disk representation with slot `db_file`; most of your own code is still interested in extracting `name()` and `age()` and not interested in `db_file()`. I think it's the transition from 'literal' to more nuanced, rather than renaming-of-slots.

S4.Rmd
mtmorgan: Since unnamed arguments to (the default) `initialize()` are taken as base classes for initialization, one pattern is

```
.Employee <- setClass(
    "Employee",
    contains = "Person",
    slots = c(
        job_title = "character",
        id = "character"
    )
)

Employee <-
    function(name = character(), age = numeric(), job_title = character())
{
    stopifnot(length(name) == length(job_title))

    id <- paste0("EID-", sample(100000, length(job_title))) # psuedo-unique
    .Employee(Person(name, age), job_title = job_title, id = id)
}
```

S4.Rmd
mtmorgan: On the 'data.frame column', S4 classes generally can't be inserted as columns, even if they have the relevant API. One could use `S4Vectors::DataFrame()` or implement a coercion method to / from data.frame, as in `library(GenomicRanges); GRanges(data.frame(gr=GRanges("1:1-10")))`

hadley: I think I'll redo the basic example here, but if S4 objects can't easily go into a data frame, why not make the objects scalars and put them into list columns? 

But maybe Person is ok for a very simple example. A more complex example (e.g. a linear model) would be more clearly scalar. I'll keep thinking about it.

S4.Rmd
hadley: I'm still not convinced by this line of reasoning, but it's clearly standard practice in the S4 community so I need to cover it.

S4.Rmd
mtmorgan: It's an efficiency issue, where one ends up iterating (over objects) instead of vectorizing (slots).

Probably a more complex model is like multiple tables in a relational data base, with the S4 object managing the relations.

Probably the idea that an S4 object is a collection of vectors is somehow encouraging 'best practice', and of course there are use cases where a list-of-scalar-objects is appropriate and performance issues are not important.

That an S4 object  can't represent a column in a data.frame is viewed (in the Bioc devel community, I guess) as a serious implementation limitation of data.frame; really, a 'column' should be defined in terms of an API that consists of a few simple methods -- `length()`, `[`, etc.; this is how `S4Vectors::DataFrame()` is implemented.

hadley: Ok, I'll emphasise that S4 objects should be vectorised where possible.

Has there been any attempt to improve R's data frames along those lines?

Should we at least make tibbles compatible with S4 objects?

mtmorgan: Making tibbles compatible with S4 objects as columns would be awesome. I defer to @lawremi and @hpages for expertise.

S4.Rmd
hpages: Hi,
It's all about encapsulation, a central concept in OO programming, not just with R and S4 objects. The idea is that the internal representation of an object is no user business so should be hidden behind a set of getters and setters. This might sound overkill for trivial objects because for these objects there will generally be a one-to-one mapping between slots and getters/setters. For more complicated objects one might want to make some slots read-only (by providing a getter but no setter), or completely hidden aka private (by providing no getter or setter). A typical private slot is one that contains cached data.

Another common situation is to have on-the-fly computed slots e.g. IRanges objects have `start()`, `end()`, and `width()` getters but only a `start` and a `width` slot (no `end` slot). The `end()` getter just does `start(x) + end(x) - 1`. Because people don't do direct slot access, they always use the getters/setters, they don't know about this and they don't care. We could change the IRanges internals anytime e.g. have a `start` and an `end` slot without breaking their code. We could also decide that we want to store the `start` and `width` vectors as Rle objects, or use other compression tricks, and add the `as.Rle` argument to the `start()`, `end()`, and `width()` getters, these getters would still return an integer vector by default (we would of course make `as.Rle` FALSE by default) and so user code wouldn't be affected.

Even in Martin's scenario of slot renaming, you're right that it would not be ideal to have the getter/setter of a slot not be named like the slot so ultimately one would want to also rename the slot. But at least by having a getter/setter you can manage a smooth transition (via .Deprecated/.Defunct cycles). If you've exposed the internals of your objects to the users and they've started to rely on them in their code, you're stuck.

H.

hadley: Yes, but generally encapsulation is about encapsulating the internal details from outside the class, not about encapsulating some details within the class.

lawremi: But there is no "within" the class. There is the class, and there are methods that do things with instances of the class. Extensibility in part comes from adding behaviors to existing data types. Therefore, the representation of the class needs to be encapsulated.

hpages: The examples I gave you are all about "encapsulating the internal details from outside the class" but maybe I don't understand what you mean exactly by this.

hadley: @hpages I thought you were responding to the thread - the text has already been changed. I understand the argument for wrapping fields with accessors, although I personally think it's overkill in many situations.

@lawremi Hmmm, that argument is somewhat compelling, but doesn't a subclass always possess the slots of its parents? I guess maybe not if you're using a virtual class as a mixin but if you're at that level of S4, I don't think you need to read this guide 😄 

Updated text on accessors at http://adv-r.hadley.nz/s4.html#accessors. It still needs rewriting, but hopefully conveys the right basic message.

hpages: As I said, it feels overkill when there is a one-to-one mapping between slots and getters/setters. But let me refine this.

It's actually overkill only if the objects are read-only. One argument for providing setters even for trivial objects is that @<- doesn't trigger object validation by default (it only checks that the type of the right value is compatible with the expected type for the slot). So the "provide accessor methods" policy could be considered overkill only for **read-only objects with a one-to-one mapping between slots and getters**. These objects are a very small subset of all the objects found in the Bioconductor ecosystem. Not providing setters for any other object will inevitably lead to the user breaking objects without realizing it.

But even the read-only objects with a one-to-one mapping between slots and getters will benefit from having getters the day you want to rename a slot and assure a smooth transition.

lawremi: @hadley The updated text looks good.  Yes, a subclass always possesses the slots of its parents, even a virtual parent. Accessors reduce coupling, subclassing increases it, so accessors become even more important as the class hierarchy becomes more complex.

@hpages Although there might be exceptions, it's more important to have a simple, consistent policy or recommended best practice: always provide and use accessors (unless you really know what you're doing).

hadley: I think root cause of my discontent is that S4 does not have private slots, and the bioconductor philosophy is to hence treat all slots as private. That means that there's no lightweight way to expose a public slot.

S4.Rmd
hadley: And then you also have to define coercion methods appropriately

S4.Rmd
lawremi: One possible compromise is the notion of properties. These would be would be accessed like slots, but the developer would still be able to encapsulate them, if necessary. Tengfei and I explored this with the objectProperties package for reference classes. We were mostly interested in observability, but the implicit encapsulation feature would be useful more broadly.

https://en.wikipedia.org/wiki/Property_(programming)

PR #851, opened 2017-08-07
Add section on active fields

thomasp85:
This PR adds a section on active fields, when and how to use them


hadley:
Thanks! I'll probably tweak it a little to be shorter, but it's a great start.

thomasp85:
Tweak away🙂

thomasp85:
I've been meaning to write a bit on deep cloning, but can't seem to find the right place to put it as cloning is mainly just discussed in passing - any preferences?

hadley:
Maybe a new (brief) section on cloning somewhere? Maybe under reference semantics?

PR #852, opened 2017-08-09
Typo

maurolepore:


hadley:
That's not a typo - it's a deliberate way to refer to the function called `class<-`

PR #855, opened 2017-08-14
Clone & deep clone

ColinFay:
Added info about deep cloning in R6, and created a full subpart about cloning.

## Copyright

### DIFF
--- a/R6.Rmd
+++ b/R6.Rmd
@@ -426,15 +426,6 @@ y1$add(10)
 y2$sum
 ```
 
-You can manually copy an R6 object by calling `$clone()`.
-
-```{r}
-y1 <- Accumulator$new() 
-y2 <- y1$clone()
-
-y1$add(10)
-y2$sum
-```
 
 There are three other less obvious consequences:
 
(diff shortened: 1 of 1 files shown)


hadley:
Could you please have a look at why the Rmd fails to compile?

R6.Rmd
```diff
@@ -447,7 +438,64 @@ There are three other less obvious consequences:
...
+original2$hadley$full_name == cloned_deep$hadley$full_name
+```
+
+When `cloned_deep$hadley` is modified, `original$hadley` is not: they no longer refer to the same object.
+
+Deep clone in R6 only copies data, methods and R6 classes. You have to implement your own methods if you're working with custom environments, reference class objects, or any other kind of object with reference semantics. 
+
+Note that you can prevent your R6 class from being clonable by adding `clonable = FALSE` in your class definition. 
+
+```{r}
```
hadley: Adding `error = TRUE` to the chunk options is a better solution if you want to show an error

ColinFay: Ow, ok ! Thanks!

PR #856, opened 2017-08-14
Set method

ColinFay:
Added info about the `set` method

## Copyright


hadley:
In what circumstance would you want to do this, instead of defining everything up front?

ColinFay:
I think one useful way to use the `set` method is when "experimenting" on an R6 class: this allows to keep track of the changes made to the class, without modifying the code that first generated the class.

hadley:
Thanks! I'll probably rewrite a bit but I really appreciate the contribution!

PR #857, opened 2017-08-14
parent exposure

ColinFay:
Added info about how to expose a class parent to his children.

## Copyright


hadley:
This also is a bit too esoteric for the book. But thanks for writing it and I hope you can use it elsewhere.

PR #858, opened 2017-08-14
share objects

ColinFay:
Share objects across R6 instances. 

## Copyright


hadley:
Thanks for writing this, but to me, it seems sufficiently esoteric to be out of scope for this chapter.

PR #859, opened 2017-08-16
Small edits to first sections of OO chapter

jennybc:
I read the intro (and a bit beyond). I think it gives a very useful overview of the OOP scene. PR is just a few copy edits and html-commented questions that popped into my mind.

A big-picture question I'll just leave here:
Is is possible to say at a very high-level what sorts of tasks each system is suited for? Like, is there any interesting anecdote re: why BioC embraced S4? When do people tend to turn to R6? I.e. why was it chosen for httr?

### DIFF
--- a/OO.Rmd
+++ b/OO.Rmd
@@ -7,7 +7,7 @@ In the following five chapters you'll learn about __object oriented programming_
 * There are multiple OOP systems to choose between. In this book, I'll focus 
   on the three that are most important in my opinion: S3, S4, and R6.
 
-* S3 and S4 come from a very different heritage to the OOP found in most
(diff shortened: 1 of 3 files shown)


hadley:
I've added a sentence to the end of each reason:


* Learning a little S3 allows your functions to return richer results 
  that have a user friendly display and programmer friendly internals.
  This is why S3 is used throughout base R.

* Investing in S4 can be helpful for building up large systems that evolve
  over many years and are written by many programmers. This is why the
  Bioconductor project uses S4 as fundamental infrastructure.
  
* Mastering R6 gives you a standard way to escape R's copy-on-modify semantics
  when needed. This is particularly important if you want to model
  real-world objects that change over time.

PR #861, opened 2017-08-22
Make na_value consistent

EdwinTh:
In the code block right before this one `na_value` is used, but now `na.value`. However, it seems they are indicating the same so changing the writing is confusing to me.


hadley:
Thanks!

PR #868, opened 2017-09-13
Re-order function types to match text

PeteHaitch:
So that the order of the function types in the code chunk is the same as in the above description


hadley:
Thanks!

PR #871, opened 2017-09-15
Fix 2 typos in OO-tradeoffs.Rmd

jimvine:
"is" to "if" and remove unnecessary "because". I assign the copyright in this edit to Hadley Wickham.


hadley:
Thanks!

PR #881, opened 2017-10-04
List + [ + NULL = list(), not list(NULL)

dkahle:
The subsetting section's table says that subsetting a list (with `[`) with `NULL` returns `list(NULL)`, but it just returns `list()`.  This corrects that.
```r
as.list(1:4)[NULL]
# list()
```


hadley:
Thanks!

PR #887, opened 2017-10-27
Typo and Call lists could be simplified

dan87134:
Arithmetic error in "Operator precedence and associativity.

As is it introduces the concept of recursion which a lot of people won't be familiar with. I think the replacement text gets across the idea that a call can contain other calls without need to introduce the term recursion... which I don't think will be used elsewhere in the book.


hadley:
Thanks!

Issue #882, opened 2017-10-11
Diagrams in Ch. 18 Expressions are not colorblind-friendly


brianwdavis:
It's very difficult to distinguish the black and purple colors with the light line-weights used. Only when zooming in (as displayed below vs as shown in context) can I easily discern the rounded corners and difference in hue. Please consider more aggressive rounding of the corners and thicker line-weights to differentiate.

![image](https://user-images.githubusercontent.com/13338560/31418228-d4aac224-ae02-11e7-91e0-4e66858695b2.png)

![image](https://user-images.githubusercontent.com/13338560/31418341-85b2348a-ae03-11e7-9e5c-b9d0dfa3128f.png)

hadley:
The hue is strikingly different to me, even at the small size in the image, and it should be even better in a printed book. Let me ask a few others.

brianwdavis:
Perhaps my poor color vision is an outlier, but it's not usually **that** bad compared to others. I'm using a (several years old) Dell LCD monitor at its native resolution. Even when I fiddle with the brightness or contrast settings, it's  difficult for me to distinguish hues with light lineweights like that diagram.

When zooming in using Chrome, I get to just about the size where I would be able to tell the colors apart at 150%, but then the text is enormous.

dan87134:
You can't depend on how an image looks on "your" monitor. There is a wide range of quality in the color management of monitors in the field.

If color or line width is supposed to convey info make both obvious, don't be subtle at all.

High res monitors often do a poor job of fine lines because, in effect, they make the lines finer than what they look like on a standard res monitor.

Printed materials use a different color gamut, CMYK, than the ones used on computer monitors. The result is an image using CMYK will appear to be muted, i.e. less bright, than one using a typical color gamut for a computer monitor like sRGB. The printed page just can't show the range of colors a computer monitor can.

Ask your printer if you can use sRGB in your images (you probably can) and then use something like Photoshop to see what they will look like after the printer converts them, (think reduces them... CMYK cannot represent all the colors on a computer screen).

That way you will end up with good images for the online version of the book.

The image below is an example of what happens when easy to recognize colors on a monitor are converted to CMYK (I used Photoshop to simulate the conversion. Photoshop can also be used to simulate what color blind people will see, those differences are really dramatic.) If you don't see a difference your monitor may need some adjustment...

Actually it also shows (the image on the right) what the on line version will look like if you publish CMYK colors on the web. Pretty dull, huh? :-)

[...]

PR #890, opened 2017-10-27
AST lines to thin for color

dan87134:
I made this as a text comment because I don't know how to "push" back a changed diagram (really new to git). It's really a change recommendation to make the diagrams more readable. As is, for all practical purposes, all the lines in the diagram are the same color. 

The lines that border the nodes are to thin. On high resolution displays like I it is very hard to distinguish the color. The lines need to be thicker, maybe make at least 12 pixels wide if you expect the color to convey information.

Also as is the lines vary in width. The box around "f" is wider than the rest. It looks like it was a screen grab instead of an export from whatever editor you are using to create the image and the "f" box was probably selected when the screen grab was done. 

All this is just IMHO of course.


hadley:
Duplicate of #882

PR #891, opened 2017-10-28
Whitespace and code structure

dan87134:
The intro makes the point that R code structured as a hierarchy. That is the key concept in this chapter. It should be used here too. 

I think it is also important to mention that whitespace in a string is retained... most readers probably have never written a parser or a grammar and given the sole example of <- may wonder why strings retain all whitespace... for example HTML does not retain all whitespace and that is something many readers would be familiar with, even fought with. 

Also it makes "usually" a more accurate quantifier... if there really was just a single exception like "x < -y" then usually wouldn't seem like the right quantifier to use.

### DIFF
--- a/Expressions.Rmd
+++ b/Expressions.Rmd
@@ -219,12 +219,18 @@ lobstr::ast((1 + 2) + 3)
 
 ### Whitespace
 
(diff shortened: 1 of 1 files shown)


hadley:
I prefer my original wording, but I'll think about it when I'm next rewriting this section.

PR #895, opened 2017-10-28
quantum??

dan87134:
I don't know how the term quantum fits here... maybe I a missing something. Heisenberg uncertainty principle? 

I think the key concept being presented here is that promises are opaque and that from R you cannot determine if an object is a promise or not, nor to determine a promise's environment. But that much detail isn't really needed.

All the reader needs to know is that an expression and an environment is all you need to evaluate an expression at a later time and that the capabilities of promises are very similar to quosures.

### DIFF
--- a/Evaluation.Rmd
+++ b/Evaluation.Rmd
@@ -14,35 +14,25 @@ library(rlang)
 
 ### Promises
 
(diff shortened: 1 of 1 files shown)


hadley:
I appreciate your enthusiasm, but this level of rewriting of in-progress chapters is not really helpful to me.

PR #898, opened 2017-10-28
subset execution environment

dan87134:
As written it say that the code in subset is executed in the data frame?? I don't think that means anything.

Proposed change clarifies what environment is being used. Also is possible that the global environment is used, it that is where subset is called.


dan87134:
I think I am looking at the example the wrong way

PR #904, opened 2017-11-05
Update S4.Rmd

ramiromagno:
Clarification on what "Again" meant when talking about `setMethod` deprecated arguments, i.e. the implicit reference to `setClass`.


hadley:
Thanks!

PR #912, opened 2017-11-19
revises to revisits for Function environments

byandell:


hadley:
Thanks! (I think that use of revises is ok in British English but not in American)

PR #915, opened 2017-11-19
Fixing some typos in chapter 19

croach:
Most of these are simple typo fixes, but I did run into a bit of a problem following some of the conversation around `expr`, `enexpr`, and `exprs`. I tried to fill in what I thought was missing in some of those cases, but there's a very good chance that what I filled in is wrong, so feel free to discard as you see fit.

As always, I assign the copyright of this contribution to Hadley Wickham.

### DIFF
--- a/Quotation.Rmd
+++ b/Quotation.Rmd
@@ -8,7 +8,7 @@ library(purrr)
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


hadley:
I'm still actively writing this section, so don't worry about bits that just don't make sense. I'll catch them on the next read through. 

Thanks!

PR #917, opened 2017-11-19
Consider beginning with more engaging material.

maurolepore:
Beginning with a dump of vocabulary and terminology seems like a way of delivering the material that is convenient for the writer but not for the reader. The risk is high that you'll lose many readers early. However, if you adapt the introduction to sufficiently motivate the readers (see my previous comment), then you'll be in a better position to present some jargon here (meaning that you may need to not change much of this sub section.)


hadley:
Can you please file all of these as a single issue with bullet points?

PR #919, opened 2017-11-19
Write for programmers of functions more than for users of functions

maurolepore:
This is a selfish suggestion -- I want to learn how S3 works to write functions with it. But I'll try to sell my point. 

For users, S3 just works. If you write for the users of S3 functions, I guess the reaction you can expect is: "Good to know, but this won't change anything of what I already do." But if you write for the programmers of S3 functions, then the reaction could be totally different: "Wow, I'll change my approach to programming right now!".

### DIFF
--- a/S3.Rmd
+++ b/S3.Rmd
@@ -5,6 +5,8 @@ source("common.R")
 library(sloop)
 ```
 
+??? For higher impact, write not for users of S3 functions but for programmers of S3 functions.
+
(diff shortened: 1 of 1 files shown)


maurolepore:
So I suggest you start with showing how you use S3. Maybe choose some functions that are very popular and not too complex. That would be something that will hook a lot of readers.

hadley:
Can you please file this as an issue, not a PR?

maurolepore:
Sure! https://github.com/hadley/adv-r/issues/921

PR #923, opened 2017-11-21
Renamed sloop to S3.Rmd

sebastian-c:
The travis builds are failing because S3.Rmd was renamed to sloop. This was probably the result of a find and replace script.


hadley:
Thanks!

PR #926, opened 2017-11-22
Assigning of `class`

EdwinTh:
Pretty sure you mean `class()<-` here, instead of `class<-()`

I can't make sense off the latter at least...


hadley:
The name of the function is `class<-` and the adv-r convention is to follow function names with `()`. It does come a bit unglued here though, but it is technically correct.

PR #929, opened 2017-11-24
Many kinds of tree diagrams

dan87134:
Why not just be declarative just to be more direct? As is the comparison really isn't adding anything. I can think of two kinds of widely used and known about "tree diagrams" in which order is important, TOC's (table of contents), and XML.  Except maybe for mathematicians working in graph theory most people assume order is important in tree diagrams.


hadley:
I think you normally think of trees like mobiles, the children can swing around.

PR #931, opened 2017-12-04
Update Subsetting.Rmd

sibusiso16:
Removed one “,”


hadley:
I think that's correct - it's a parenthetical around "finally"

PR #934, opened 2017-12-05
Update the `lapply` example in the Lazy Evaluation section (fixes #837)

### LINKED ISSUE #837: First lazy evaluation example is unclear
image

This example does not convey the lazy evaluation, e.g. what a reader will attempt is following:

```r
> add <- function(x) {
+     function(y) x + y
+ }
> adders <- lapply(1:10, add)
> adders[[1]](10)
[1] 11
> adders[[10]](10)
[1] 20
> adders[[1]](1)
[1] 2
> adders[[10]](1)
[1] 11
```
Which in no way differs from the example meant to alleviate the issue:
```r
> add <- function(x) {
+     force(x)
+     function(y) x + y
+ }
> adders2 <- lapply(1:10, add)
> adders2[[1]](10)
[1] 11
> adders2[[10]](10)
[1] 20
> adders2[[1]](1)
[1] 2
> adders2[[10]](1)
[1] 11
```

bborgesr:
[R 3.2.0 modified the behavior of lapply](https://stat.ethz.ch/pipermail/r-announce/2015/000583.html) to force its functions' arguments by default:

> Higher order functions such as the apply functions and Reduce() now force arguments to the functions they apply in order to eliminate undesirable interactions between lazy evaluation and variable capture in closures.

So this section needed to be brought up to date. I tried to change as little as possible, while still illustrating the potential problem and also mentioning that `lapply()` now does the same thing @hadley is recommending (forcing the functions' arguments).

Fixes #837.

### DIFF
--- a/Functions.Rmd
+++ b/Functions.Rmd
@@ -546,7 +546,11 @@ f <- function(x) {
 f(stop("This is an error!"))
 ```
 
-This is important when creating closures with `lapply()` or a loop:
+The `apply` functions underwent this same change in [R 3.2.0](https://stat.ethz.ch/pipermail/r-announce/2015/000583.html): 
+
+> Higher order functions such as the apply functions and Reduce() now force arguments to the functions they apply in order to eliminate undesirable interactions between lazy evaluation and variable capture in closures.
+
+So, as of R 3.2.0 (but not older versions), you can safely do:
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #940, opened 2017-12-18
typo

jimmyliu2017:


hadley:
Thanks for the fix, but unfortunately I rewrote Environments so I can't merge it.

PR #948, opened 2018-01-09
Fix bug due to !! precedence

eaurele:
Unless !! precedence has changed (I use rlang version 0.1.6), I cannot seem to reproduce starting with `map2(coef_sym, coefs, ~ expr((!!.x * !!.y)))`.

I propose the change to UQ() to produce the expected results


hadley:
Precedence has changed, hence the message at the top of the doc.

PR #949, opened 2018-01-12
Update Rcpp.Rmd

dongzhuoer:
the original code is `p * if (female) 1.25 else 0.75` not `p + if (female) 1.25 else 0.75`


hadley:
Thanks!

Issue #932, opened 2017-12-05
OOP S3: when to use attributes vs subclasses


maxheld83:
I'm having a hard time deciding whether to use fields or subclasses for some use cases and would appreciate any heuristics.

For example, I might have a vector of items (for some survey) of class `myItems`.
Let's say that I want to enable text-based and image-based items.

Doing this via **subclasses** would prepend something like `class = c("myItemsText", "myItems")`, and I'd then add appropriate methods for the `myItemsText` and `myItemsImage` subclasses.

But I realise this could also be done simply by adding another **attribute** (to a vector-style class), such as `type = "image"` / `type = "text"`. In this case, I'd have only *one* method (for `myItemsText`) and simply use `switch` on `type` for some of that code.

I should note that there *is* some code base overlap between text and image items, so two entirely separate classes would not make sense.

I would appreciate any guidance on this issue, and perhaps a short addition to the book.

---

**Justification**:

Yeah, I know, this seems slightly OCD, and there's probably no hard and fast *technical* reason to do one or the other, and in any case, it *depends*.
Still, I find these kinds of rules/heuristics immensely helpful just to free up cognitive load for *actual* programming. 
This, to me, is one of the great benefits of Hadleys [adv-r S3 chapter](https://adv-r.hadley.nz/s3.html), a lot of his other work and the tidyverse in general: it sets up one, reasonably thought-through script for how to do things, so I can stop re-factoring myself into a frenzy and focus on my substantive problem at hand.

Obviously, a better heuristic is preferable over a bad heuristic, hence my question.

maxheld83:
**candidate heuristic 1**:

- when the parent class (`myItems`) never makes sense / must never occur without some subclass (`myItemsImage`) -> use attributes, not subclasses (because it's not a "real" parent class).
- otherwise: use `subclasses`.

However, I'm not sure, whether this isn't merely a semantic distinction. 
For example, one might argue just as well, that the class `factor` *only* makes sense with the information `ordered = TRUE` or `ordered = FALSE`, which would imply that `ordered` should *not* be a subclass.

maxheld83:
**candidate heuristic 2:

- when sub-thing needs additional arguments, use a `subclass` (such as when `myItemsText` has a `language` attribute, but `myItemsImage` does not).
- otherwise: use attributes. 

Apparent advantage might be that this cleans up the arguments for the validators, constructors and helpers.

Still not sure though.

maxheld83:
[...]

PR #955, opened 2018-02-04
add heuristics for when to use subclass vs attributes as per #932

maxheld83:


hadley:
Given the revised focus of this chapter, I now think that this is out of scope. But please hold it in reserve for when I start working on the "tidy tools" book!

PR #959, opened 2018-02-13
Change "?? bytes" to "variable number of bytes"

miguelmorin:
I was confused when I read this chapter because "??" is TeX for "reference not found" or "file not found".


hadley:
Thanks!

PR #967, opened 2018-02-23
correct R6 active fields example

ksens:
active values definition referenced `name` where it should have been referencing the argument `value`


hadley:
Thanks!

PR #973, opened 2018-03-10
updated name of rlang-function

henningsway:
- "In rlang 0.2.0, `lang_standardise()` was soft-deprecated and renamed to `call_standardise()`."


hadley:
Thanks!

PR #982, opened 2018-03-11
I believe the existing R chunk code is wrong based on prose

daattali:


hadley:
Thanks - I tweak the wording slightly differently

PR #984, opened 2018-03-11
Show both data.frames in last diagram

EmilHvitfeldt:
When seeing the last diagram in "4.3.4 Data frames" it is initially hard to see that the diagram is different then the first diagram in the same section. Showing them side by side might help understanding. Since it worked quite well in section "4.3 Copy-on-modify".


hadley:
Thanks! In the future, just file issues for image changes because I need to change the source.

PR #987, opened 2018-03-11
set cache=FALSE to fix section 4.2 get_addr inconsistency

cortinah:


hadley:
Thanks, but I think using `dependson` is a cleaner fix.

PR #988, opened 2018-03-12
Update link to local groups

csgillespie:
About a year ago, we started a GitHub repo of user groups (https://github.com/jumpingrivers/meetingsR/) The Revo blog (http://blog.revolutionanalytics.com/local-r-groups.html) points to the page.


hadley:
Thanks!

PR #1000, opened 2018-03-26
Update Performance.md

alexperrone:
Previous version referred to RC incorrectly. It said for "10,000 dispatch, 10,000 ns" when it should have said "RC dispatch, 10,000 ns." This was corrected.


hadley:
Thanks!

PR #1001, opened 2018-03-27
associative rule for addition

shabbybanks:
what was described was the commutative rule, not the associative rule, for regular addition.


hadley:
Thanks!

PR #1010, opened 2018-04-08
Update "Rcpp is smoking fast..." link

malcolmbarrett:
The link for "Rcpp is smoking fast for agent-based models in data frames" is dead, and it's now being rehosted at https://gweissman.github.io/babelgraph/blog/2017/06/15/rcpp-is-smoking-fast-for-agent-based-models-in-data-frames.html. I updated the link, although I understand if keeping the original link is more for citation than linking.


hadley:
Thanks!

PR #1014, opened 2018-04-16
Without `data` argument, an error occurs

taekyunk:
Without `, data = data`, the following fails with an error as shown.

```
> subset3(iris, Species == "setosa")
 Error in ~Species == "setosa" : object 'Species' not found 
```


hadley:
Ooops, thanks!

PR #1015, opened 2018-04-16
Parent environment

nietzschetmh:
Hi, In line 154, 
e2b <- env(e2a, a = 1, b = 2, c = 3) 
will give me error like this:
Error: !length(data) || is_named(data) is not TRUE

According to the help page of env{rlang}, env() always creates a child of the current environment.
So if we want to create a parent environment, should we change it to e2b <- child_env(e2a, a = 1, b = 2, c = 3)  ?

### DIFF
--- a/Environments.Rmd
+++ b/Environments.Rmd
@@ -151,7 +151,7 @@ You can set the parent environment by supplying an unnamed argument to `env()`. 
 
 ```{r}
 e2a <- env(d = 4, e = 5)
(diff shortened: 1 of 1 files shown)


hadley:
Do you have the dev version of rlang?

nietzschetmh:
rlang package version is 0.2.0

hadley:
You need the development version

PR #1017, opened 2018-04-19
Use nested function to demonstrate message2error.

jdblischak:
When demonstrating the behavior of `withCallingHandlers()` via `message2error()`, a function `f()` is defined but never used. I updated the example to use this function. A later exercise (currently in [Conditions.Rmd](https://github.com/hadley/adv-r/blob/a36fa316a6040f8d27e7dff706b2348be250afd6/Conditions.Rmd)) compares the difference in the call stacks when `message2error()` is implemented with `withCallingHandlers()` versus `tryCatch()`. The difference is more noticeable when the message is produced within a nested function, i.e. the first entry `message2error(g())` in the traceback after using `tryCatch()` appears rather informative since `g()` itself contains the message.

### DIFF
--- a/Debugging.Rmd
+++ b/Debugging.Rmd
@@ -309,23 +309,26 @@ There are other ways for a function to fail apart from throwing an error or retu
 
     f <- function() g()
     g <- function() message("Hi!")
-    g()
-    # Hi!   
-    message2error(g())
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1021, opened 2018-04-25
Swap order matrix arguments

prosoitos:
This is less confusing as it is consistent:

- with the default matrix argument order

`matrix(data = NA, nrow = 1, ncol = 1, byrow = FALSE,
       dimnames = NULL)`

- with the corresponding vectorized form

`matrix(1:6, c(2, 3))`


hadley:
Thanks!

prosoitos:
My pleasure!

PR #1022, opened 2018-04-25
Fix typos

prosoitos:
- in Preface.Rmd

Fixed "this goal of this package" to "the goal..."

### DIFF
--- a/Preface.Rmd
+++ b/Preface.Rmd
@@ -6,7 +6,7 @@ The 2nd edition has been published in colour, which as well as improving the syn
 
 ## rlang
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

prosoitos:
Oops! I just pushed a 3rd commit to my `typos` branch. I didn't expect the PR to be accepted so quickly and I didn't realize you had already merged it! :smile: 

I guess I can create a new branch and submit a new PR with that last change.

Also: I realize that this is a draft version of a work in progress. So, is typo correction useful or just an annoyance at this point? If it is too early for that level of detail, I can refrain from fixing every missing article and only focus on content.

hadley:
It's useful for the chapters that are a bit more polished — if there's a bunch of words obviously missing (like the preface, you can usually just ignore and I'll find eventually)

PR #1028, opened 2018-04-26
Fix a few typos in Names-values.Rmd

prosoitos:


Names-values.Rmd
```diff
@@ -301,7 +301,7 @@ knitr::include_graphics("diagrams/name-value/d-modify-r.png", dpi = 300)
 ### Character vectors
 \index{string pool}
 
-The final place that R uses references is in character vectors. In the previous chapter, we drew character vectors like this:
+The final place where R uses references is in character vectors. In the previous section, we drew vectors like this:
```
prosoitos: I misunderstood this because the "previous chapter" has now become the "next chapter". I thought this whole sentence was bizarre and I thought that it might refer to the vectors drawn in the previous section. Hence my edits.

Reading the "4. Vectors" chapter, I realized that the sentence initially made sense and simply was not working anymore because the place of "3. Names and values" must have changed at some point, since it is an additional chapter. So this section needs editing, but not the way I did...

prosoitos:
Thank you for your feed-back!

Here are my replies to your comments:

2. & 3.: you're right! my bad. I will revert those edits.
5: I guess both work, but I will happily revert.

4. I guess the sentence can be understood in 2 ways. If you "translate" `?` by "help file of", I agree with you that the sentence would be better with "in" (as in: see the list **in** "the help file of" Reserved). Or maybe you understand it as: see the list **in** Reserved, ignoring the `?` that you consider as a side information on how to see `Reserved` that should not impact the sentence meaning or structure.

The way I took it is with a literal reading of the expression `?Reserved`. As in: see the list **with** `?Reserved` (= you can find that list by typing `?Reserved`. As we would say, see the structure of df **with** `str(df)`).

So I guess both work, but I really prefer "with". Now that I have showed you how I read the sentence, please let me know if you see it differently or still prefer "in".

6. (but/instead): it is more than stylistic. "Instead" introduces an alternative (see for instance https://wikidiff.com/instead/but). It isn't about an alternative here. (You won't use these crazy names, **but** you still need to know them because...). Needing to know the words is not an alternative to using the crazy names. "Instead" would work well if the situation was: Don't use such crazy names. **Instead** use this or that name).

7. I am not sure whether you saw my 3rd commit: I first made an edit, replacing "chapter" by "section" because the sentence made no sense to me. But I later realized what the problem was: it is in Chapter 4 Vectors" that he drew diagrams that way. And Chapter 4 is after. I imagine that the sentence was written when this novel chapter (chapter 3), which was not in the 1st edition, was placed after chapter 4. Then I imagine that the order of the chapters was changed. And that is why the sentence made no sense anymore. In my final edit, I removed that sentence altogether (saying that he will represents them as if... in the rest of the book is sufficient since there was no such representation before).

jtr13:
4. (with vs  in) I was indeed thinking of it as "in the help file of"

Before going on to other points, a metacomment: it would be helpful to have clearer instructions from the author what kind of edits he would like to see in pull requests.  In the absence of any such instructions, my approach is to only make pull requests for very obvious typos rather than do a line by line edit. In a case like this, where "in" works, I would just leave it be. Again, that's just a guess on what would be most helpful to the author. 

6. Agree that "but" is a better choice than "instead," but it doesn't seem clearcut enough to bother with

7. Correct, it was hard to follow what was going on. The changes aren't always prominent in the diff. The only thing that was clearly wrong was the reference to the previous chapter. Again, the minimalist approach.

jtr13:
Second that... @hadley: it would be really helpful to have a blurb in the book explaining what kind of edits would be helpful to you.  I'm in guess mode.

hadley:
Before making big changes, it's best to check with me first. I can easily accept typos and minor fixes without further discussion, but if you want to change wording it's better to start small so I can give you feedback on the voice/tone/style.

Unfortunately many of these changes subtly change the intent of the phrasing so I can't accept them.

prosoitos:
OK. Thank you for the feed-back.

PR #1031, opened 2018-04-27
Fixes chapter "5. Subsetting"

prosoitos:


Subsetting.Rmd
```diff
@@ -108,23 +108,24 @@ There are five things that you can use to subset a vector:
...
+       x[c(TRUE, FALSE, TRUE, FALSE)]
+            ```
 
     A missing value in the index always yields a missing value in the output:
 
-    ```{r}
+    ```{r}																										
     x[c(TRUE, TRUE, NA, FALSE)]
-    ```
+        ```
```
prosoitos: Sorry for this formatting non-sense: I don't use rmarkdown (I use org-babel) and some of my edits messed up the rmarkdown formatting in emacs polymode, that I then tried to fix.

jtr13:
Looked it up since I wasn't sure about editing "being subsetted" to "being subset" and it turns out that the use of "subsetted" in this very chapter is the subject of a whole discussion on Stack Exchange https://english.stackexchange.com/questions/297323/simple-past-of-the-verb-subset No clear answer :-)

prosoitos:
I know: googling it lead me to the same thread :smile: Truth is, the verb doesn't exist. So I guess, in that case, one can do what one wants. But "subset" would be more consistent with what happens in similar situations than "subsetted" since the preterite and part participles of "set" are both "set". For instance, the preterite and past participle of "reset" are both "reset" (not "resetted").

jtr13:
Agree. Imho, subset is more logical but subsetted sounds better and more clearly communicates that an action was performed.

prosoitos:
Thanks @hadley for bearing with my clumsy noobs attempts while I am learning to contribute in better ways :stuck_out_tongue:

PR #1033, opened 2018-04-27
Typo: predictor -> response variable

taekyunk:
Fixing a typo


hadley:
I merged another PR that also fixed this problems - but thanks!

PR #1034, opened 2018-04-28
Update Conditions.Rmd

mattfidler:
Add a note about `str` and `message` in Rstudio.

I'm unsure if this is a Rstudio bug instead.


hadley:
You should always use `cat()` in a `str()` method

mattfidler:
Correct.  Do you want to point it out?

PR #1035, opened 2018-04-29
A chunk is run that shouldn't

maurolepore:
It seems that this chunk is meant to not run, but it does:


hadley:
The problem was `result` should have been `results`

PR #1043, opened 2018-05-07
Fixed two typos related to warnings

cortinah:
1) file() warns if the file _does not_ exist
2) when you try _to_ read, not try _and_ read  (style)


hadley:
For now, it's probably better now to make any changes to this chapter since I'm so actively working on it. Thanks for the suggestions though!

PR #1045, opened 2018-05-08
Small typo fix in Quotation.Rmd

osorensen:
From *... that different only slightly ...* to *... that differ only slightly ...*.


hadley:
Thanks!

PR #1047, opened 2018-05-12
Add attributes(a) before examples

blmayer:
Made attributes examples clearer since they were returning NULL and NULL, not giving the correct impression of loosing attributes.


hadley:
Thanks!

PR #1048, opened 2018-05-12
Maybe explain: Each scalar-style object represents a single “value”

maurolepore:
> Each scalar-style object represents a single “value”, and are built on top of named lists. 

I find this confusiong. How is that an object based on a list represents a single value? 
The way I interpret this -- and with a lot of effort -- is this: The "singlarity" of this style comes from the fact that all the fields of this object are contained in a _single_ package, the list. Maybe this needs a little more explanation.


hadley:
Can you please file this as an issue? It is not helpful as a PR.

PR #1051, opened 2018-05-13
Do you mean `.data` (dot-data)?

maurolepore:
Also, is `!!var`  necessary?


hadley:
`!!var` is necessary in case df contains a variable called `var`

PR #1063, opened 2018-05-16
Added missing dot to "data"

osorensen:
The example function threshold_var1 does not run the way it is defined now. Added "." in front of data, and changed the explanation of the code accordingly.


hadley:
Thanks!

PR #1066, opened 2018-05-17
Clarification of Evaluation example

osorensen:
I wonder if this example would be clearer for the reader if `env` was defined, even though it does not matter for the result of this function. Then  `eval_bare(x + y, env(x = 1000))` gives 12 while `eval_bare(expr(x + y), env(x = 1000))` gives 1002. 

As it is now, `eval_bare(expr(x + y), env = env)` fails, since `env` is a function, not an environment.


hadley:
Looks good - thanks!

PR #1078, opened 2018-05-23
Unnecessary arguments to abort()

osorensen:
I think the arguments `arg`, `must`, and `not` are unnecessary, since the whole string is glued together into `msg`. It does not make any difference for the output of the function, though.

``` r
# I use rlang 0.2.0.9001 and R 3.5.0.
library(rlang)

abort_bad_argument <- function(arg, must, not = NULL) {
  msg <- glue::glue("`{arg}` must {must}")
  if (!is.null(not)) {
    msg <- glue::glue("{msg}; not {not}")
  }
  
  abort("bad_argument_error", 
        message = msg # arg, must, and not removed
  )
}


log <- function(x, base = exp(1)) {
  if (!is.numeric(x)) {
    abort_bad_argument("x", must = "be numeric", not = typeof(x))
  }
  if (!is.numeric(base)) {
    abort_bad_argument("base", must = "be numeric", not = typeof(base))
  }
  
  base::log(x, base = base)
}


log(letters)
#> Error: `x` must be numeric; not character
log(1:10, base = letters)
#> Error: `base` must be numeric; not character
```

### DIFF
--- a/Conditions.Rmd
+++ b/Conditions.Rmd
@@ -44,24 +44,6 @@ Want to skip this chapter? Go for it, if you can answer the questions below. Fin
 ```{r setup}
 library(rlang)
 
-cnd_signal <- function(cnd) {
-  switch(cnd_type(cnd), 
-    error = cnd_abort(cnd),
-    warning = cnd_warn(cnd),
-    message = cnd_inform(cnd),
-    condition = rlang::cnd_signal(cnd, .mufflable = TRUE)
-  )
-}
-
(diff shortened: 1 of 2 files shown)


hadley:
Unfortunately there's not much point in commenting on this chapter since it's still being written, and the tools in rlang are evolving too.

osorensen:
Ok, thanks. I learned a lot from reading it anyway!

PR #1089, opened 2018-06-02
Typo in getting and setting

jonthegeek:
Is this level of typo editing helpful?


hadley:
Yes, but it would be even more useful if you fixed multiple typos per PR :)

jonthegeek:
I thought that might be the case! I was fixing them as I found them at first, but it started to feel tedious (for you as much as me), so I stopped. I need to re-read those chapters anyway, so I'll do that re-read in a text editor.

PR #1100, opened 2018-06-25
found some possible mistakes

jhrcook:
For some of the function calls, I believe the output was unexpected.


hadley:
Can you please file issues instead of PRs?

PR #1105, opened 2018-06-28
Update Functions.Rmd

sedaghatfar:
Spelling 

encounted -> encountered
use name -> use a name
thingswill -> things will
Vaue -> value
Propety -> property


hadley:
Thanks!

PR #1112, opened 2018-07-06
Update broken link to SICP full text

hamedbh:
Replaced the link with one that works as of 2018-07-06. 


hadley:
Thanks!

PR #1114, opened 2018-07-06
Functions:Introduction:Quiz

lindbrook:
Link in the middle of the sentence?

### DIFF
--- a/Functions.Rmd
+++ b/Functions.Rmd
@@ -11,7 +11,7 @@ If you're reading this book, I expect that you've already created many R functio
 
 ### Quiz {-}
 
-Answer the following questions to see if you can safely skip this chapter. You can find the [answers][\@ref(function-answer)] at the end of the chapter.
+Answer the following questions to see if you can safely skip this chapter. You can find the answers in \@ref(function-answer).
 
 1.  What are the three components of a function?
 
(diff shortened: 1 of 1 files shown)


Functions.Rmd
```diff
@@ -11,7 +11,7 @@ If you're reading this book, I expect that you've already created many R functio
 
 ### Quiz {-}
 
-Answer the following questions to see if you can safely skip this chapter. You can find the answers at the end of the chapter in [answers](#function-answers).
+Answer the following questions to see if you can safely skip this chapter. You can find the [answers](#function-answers) at the end of the chapter.
```
hadley: We need to rewrite these to be (e.g.) in `Section \@ref(function-answer)` so they look the same in online and printed books.

lindbrook:
Trying to avoid using "answers" twice. Not sure if this works stylistically or syntactically.

hadley:
Maybe just `You can find the answers in \@ref(function-answer)?`. 

Alternatively, now that we have footnotes, we could put the answer to each question in its own footnote?

lindbrook:
Probably better in a separate section (at end of chapter?). Less peaking.

hadley:
BTW for this edition of the book (since I'm now using bookdown), I'm slowly moving all cross-links to the `Section \@(ref)` form

PR #1128, opened 2018-07-19
Added `of`

isteves:
This is one of those grammar things that makes me doubt myself, because I think it might be fine to drop `of` in speech.  In any case, I think it reads a bit better with it.


hadley:
Thanks, it's definitely better!

PR #1132, opened 2018-07-21
Functions:Primitive functions

lindbrook:


Functions.Rmd
```diff
@@ -198,7 +198,7 @@ body(sum)
 environment(sum)
 ```
 
-Primitive functions are only found in the base package. Primitive functions have certain performance advantages, but come at the cost of increased implementation complexity. Primitive functions are harder to write, and R-core generally avoids creating them unless there is no other option.
+Primitive functions are only found in the base package. While they have certain performance advantages, this benefit comes at a price. They are harder to write. So R-core generally avoids creating them unless there is no other option.
```
hadley: Maybe "at a price:" ? Otherwise the next sentence seems a bit short.

lindbrook: Always go with your first thought.

hadley:
Thanks!

PR #1134, opened 2018-07-23
Fix typos in Vectors.Rmd

jemus42:
There were letters missing, but I think I found them and am now returning them home.

Even if I'm confident he would have found them on his own ¯\\\_(ツ)_/¯


hadley:
Thanks!

PR #1135, opened 2018-07-25
Wordsmith the MLE/function factory section

jennybc:


Function-factories.Rmd
```diff
@@ -191,23 +188,23 @@ lprob_poisson <- function(lambda, x) {
...
 ```{r}
 lprob_poisson(10, x1)
 lprob_poisson(20, x1)
 lprob_poisson(30, x1)
 ```
 
-The likelihood is the probability function, just seen through a different lens: we want to fix `x` and find the `lambda` that yields the highest result. In R, we can make this distinction more clear by using a function factory. We provide `x` and generate a function with a single parameter, `lambda`:
+The key idea of maximum likelihood is to think about the probability function in a different way. So far we've been thinking of `lambda` as fixed and known and the function tells us the probability of getting different `x` values. But in real-life, we observe the `x` and it is `lambda` that is unknown. The likelihood is the probability function, seen through this lens: we want to find the `lambda` that makes the observed `x` the "most likely". That is, given `x`, what value of `lambda` gives us the highest value of `lprob_poisson()`? 
+
+In R, we can make this change in perspective more clear by using a function factory. We provide `x` and generate a function with a single parameter, `lambda`:
```
hadley: Do you think I should mention the equivalent math-y notation?

hadley:
Thanks!

Function-factories.Rmd
jennybc: That seems like a "nice to do" but not a "must do". You'd have to settle on more notation, i.e. for the (log) likelihood, without getting to use it very much.

PR #1141, opened 2018-08-02
Remove warning about deparsing !!

lionel-:
Since it is no longer an issue with R 3.5.

And explain why using !! outside of quasiquoting functions causes trouble


Quotation.Rmd
```diff
@@ -376,20 +376,26 @@ So far we have acted as if `!!` and `!!!` are regular prefix operators like `+`
...
-```{r}
-x <- quote(!!x + !!y)
-deparse(x)
+```{r, error = TRUE}
+x <- quote(variable)
+!!x
 ```
 
-Although the R parser can distinguish between `!(x)` and `!x`, the deparser currently does not. You are most likely to see this when printing the source for a function in another package, where the source references have been lost. `rlang::expr_deparse()` works around this problem if you need to manually deparse an expression, but often this does not help because the deparsing occurs outside of your control, as during debugging.
+However be extra careful when unquoting numeric values that can be negated silently:
```
hadley: 3.5 still feels like a bit recent to drop altogether. Maybe mention in a footnote?

lionel-: Done in last commit.

PR #1142, opened 2018-08-03
Grammar corrected

setempler:
Not sure if the expression was meant to be ‚as is’, since I am no native speaker.

Duckduckgoing "all we need to turn is turn" did not yield any result, giving me confidence it is meant to be as proposed here.

Please excuse inconvenience, if so, and thank you for the nice read!


hadley:
Thanks!

PR #1146, opened 2018-08-06
Change "try and" into "try to"

kwstat:
If you accept this, I'll submit a few more similar changes.

Commentary on "try and":
http://www.macmillandictionaryblog.com/try-to-get-over-try-and


hadley:
Looks good, thanks!

PR #1157, opened 2018-08-13
more Functions:Forcing evaluation

lindbrook:


Functions.Rmd
```diff
@@ -474,9 +474,9 @@ h02 <- function(x) {
 h02(stop("This is an error!"))
 ```
 
-It is usually not necessary to force evaluation. It's needed primarily for certain functional programming techniques which we'll cover in detail in [function operators]. Here, I want to show you the basic issue.
+It's usually not necessary to force evaluation. It's needed primarily for a functional programming techniques we'll cover in detail in [function operators]. Here, I want to show what the basic issue is.
```
hadley: techniques -> technique

lindbrook: Just to be sure. Are function operators the only functional programming technique that uses force evaluation?

hadley: Well, function factories also should, but it's not quite as important

PR #1168, opened 2018-08-21
Switch users' to singular possessive

batpigandme:
You use "the user" (singular) as the section continues, so I think it makes more sense this way.

Bonus fun fact: 'twas the only place you used "users'" in the whole book. 🕴


hadley:
Oops

PR #1174, opened 2018-08-30
Add RStudio Community to "getting help"

tbradley1013:
I added references to RStudio Community in the "Getting Help" section and added two sentences about using the `reprex` package to the bullet point about creating a reproducible example.


tbradley1013:
I don't think this is considered a "significant" change, but just in case:

hadley:
Thank you!

PR #1177, opened 2018-09-04
Remove extra hashtag in ref

isteves:
Currently showing up as:

image


hadley:
Thanks @isteves! (FWIW you can just fix the problem - no need to explain what it is as I can usually spot it when I see the diff)

isteves:
Sounds good.  Figured it was better to err on the side of being overly clear!

PR #1180, opened 2018-09-14
fix a typo in Answers

ktang:
Changed Line 717 from 
1.  It occupies about 4 MB.
to
1.  It occupies about 8 MB.


hadley:
Thanks! For future reference, you don't need to describe the change as I'll see it in the diff

PR #1182, opened 2018-09-14
Changing 'mental' to 'mental model'

gvwilson:


hadley:
Thanks, although obviously I still have to rewrite this whole chapter.

PR #1196, opened 2018-09-16
Update new_subclass constructor in S3 chapter

drtjc:
Change `z` argument to be named argument: `z = z`


drtjc:
realized that this change also needs to be made for code when subclass constructor allows its own subclass. will make change and send PR

PR #1203, opened 2018-09-18
Add missing words in FP intro

batpigandme:
n.b. I added the _is_ in https://github.com/batpigandme/adv-r/commit/a7cf4b0a25fcf1f972104926b290917a66d3d888#diff-ed8b1340d157a79b8555ae20e0459f2fR12
> A functional style tends to create functions that can easily be analysed in isolation (i.e. using only local information), and hence _**is**_ often much easier to automatically optimise or parallelise.

It's possible that the _is_ should be _are_ and the subject is _functions_ and not _A functional style_, but either way, I think there should be a verb there.


hadley:
Thanks - I'll read through again today

PR #1210, opened 2018-09-20
deleting "for"

ktang:


hadley:
I'm pretty sure this is correct

PR #1228, opened 2018-09-26
Functions:Functions vs. variables

lindbrook:


Functions.Rmd
```diff
@@ -421,7 +421,7 @@ g08 <- function() {
 g08()
 ```
 
-The rule gets a little more complicated when a name is bound to a function and a non-function in different environments. When you use a name in a function call, R will ignore non-function objects while looking for that value. For example, here `g09` takes on two different values:
+The rule gets a little more complicated when a name is bound to a function and a non-function that lie in different environments. When you use a name in a function call, R will ignore non-function objects while looking for that value. For example, here `g09` takes on two different values:
```
hadley: lie or live?

lindbrook: how about "reside"? Also, when looking for a value, is it that R gives priority to function objects rather than simply ignores non-function objects?

hadley: Reside sounds good. 

It is ignoring - obviously you can't have a function and non-function bound to the same name in the same environment, so when looking for names that must be functions, R skips past environments where the name is bound to a non-function.

Maybe this needs a picture?

lindbrook: Right. But if there are no function objects with that name, won't R then start to look at environments with that name bound to non-function objects?

hadley: The order in which environments are considered stays the same, it's just that some might be skipped if they contain a binding to non-function

hadley:
Thanks!

PR #1230, opened 2018-09-26
Functions:Dynamic lookup

lindbrook:


Functions.Rmd
```diff
@@ -458,7 +458,7 @@ You might be surprised that `g11()` always returns the same value. This happens
 
 ### Dynamic lookup
 
-Lexical scoping determines where to look for values, not when to look for them. R looks for values when the function is run, not when it's created. This means that the output of a function can differ depending on objects outside its environment: 
+Lexical scoping determines where, not when to look for values. R looks for values when the function is run, not when the function is created. Together, these two properties tell us that the output of a function can differ depending on the objects outside the function's environment:
```
hadley: Maybe "where, but not when, ..."?

lindbrook:
what do "both of these approaches" in the second to last sentence refer to?

Functions.Rmd
```diff
@@ -469,23 +469,23 @@ x <- 20
...
-Another way to solve the problem would be to manually change the environment of the function to the `emptyenv()`, an environment which contains nothing:
+Another way to solve the problem is to manually change the function's environment to the `emptyenv()`, an environment which contains nothing:
 
 ```{r, error = TRUE}
 environment(g12) <- emptyenv()
 g12()
 ```
 
-Both of these approaches reveal why this undesirable behaviour exists: R relies on lexical scoping to find _everything_, even the `+` operator. This provides a rather beautiful simplicity to R's scoping rules.
+Both of these approaches reveal why this seemingly undesirable behaviour exists: R relies on lexical scoping to find _everything_, even the `+` operator. This gives R's scoping rules a rather beautiful simplicity.
```
hadley: I think approaches should be examples?

lindbrook: Re: "... R relies on lexical scoping to find _everything_, even the `+` operator"

To illustrate "_everything_", it might be nice to show a contrast. Something along the lines of: "from the humble `+` operator to [something less humble] ".  Suggestion for the latter?

hadley: How about "from the obvious, like `mean()`, to the less obvious, like `+` and even `{`" ?

hadley:
Thanks!

PR #1234, opened 2018-09-27
Meta fixes

hammer:


hammer:
I figured you probably wanted to link to the `Quasiquotation` chapter heading rather than the `Quotation` section heading, and I think I fixed the `Translation` link by using the explicit chapter ID, but did not test in `bookdown` to confirm.

hadley:
Thanks - these will all eventually get translated to `Section \@ref(blah)` so feel free to skip and I'll eventually fix.

PR #1244, opened 2018-10-02
Replace `lang()` with `call2()`

batpigandme:
* Since `rlang::lang()` is soft-deprecated and gives warning instructing to switch to `call2()`


hadley:
Thanks!

PR #1249, opened 2018-10-02
Hyphenate first-class environments

batpigandme:
I left `first class function(s)`, but modified to compound modifier for env't(s). (IMHO, first-class functions should get a hyphen, too. Just LMK). 😳


hadley:
Yeah, can you please apply to all instances of "first class" that should be "first-class"

PR #1257, opened 2018-10-05
Comments on R6 chapter

gvwilson:
This was a _lot_ easier to follow than the S4 chapter, and I don't think that's just because R6/RC is class-based rather than generic-based - I think it's because single inheritance allows you to avoid a boatload of emoji diagrams. As I mention twice in the chapter, I think S3 -> R6 -> S4 will be a much gentler reading order than historical order.


hadley:
Rearranging the chapters is a great idea — there's no need to recapitulate history here.

PR #1263, opened 2018-10-11
Functions:Missing arguments

lindbrook:


Functions.Rmd
```diff
@@ -683,9 +683,9 @@ str(h09(10))
 args(sample)
 ```
  
-It looks like both `x` and `size` are required, but in fact `sample()` uses `missing()` to provide a default for `size` if it's not supplied. If I was to rewrite sample myself[^sample], I'd use an explicit `NULL` to indicate that `size` can be supplied, but it's not required:
+It looks like both `x` and `size` are required, but in fact `sample()` uses `missing()` to provide a default for `size` if it's not supplied. If I were to rewrite sample myself[^sample], I'd use an explicit `NULL` to indicate that `size` is not required but can be supplied:
 
-[^sample]: Note that this only implements one way of calling `sample()`: you can also call it with a single integer, like `sample(10)`. This unfortunately makes `sample()` prone to silent errors in situations like `sample(x[i])`.
+[^sample]: Note that this only one way to implement a call to `sample()`: you can also call it by using a single integer, like `sample(10)`. This unfortunately makes `sample()` prone to silent errors in situations like `sample(x[i])`
```
hadley: I think this loses the original intent; i.e. there are multiple ways to call `base::sample()`, and I only implement one.

Functions.Rmd
```diff
@@ -697,7 +697,7 @@ sample <- function(x, size = NULL, replace = FALSE, prob = NULL) {
 }
 ```
 
-You can make that pattern even simpler with a small helper. The infix `%||%` function uses the LHS if it's not null, otherwise it uses the RHS:
+You can simplify this task with the help of the infix `%||%` function, which uses the LHS if it's not `NULL` and the RHS otherwise:
```
hadley: I think pattern is important here

Functions.Rmd
hadley: That's not quite right either. Maybe we should just kill this footnote altogether? It's not that important a point here.

Functions.Rmd
hadley: The pattern is `argument %||% default_value`, and you might use that pattern in many different functions.

Functions.Rmd
lindbrook: ok. will drop it.

PR #1264, opened 2018-10-12
Vectors

gvwilson:


Vectors.Rmd
```diff
@@ -237,6 +255,7 @@ You can name a vector in three ways:
 x <- c(a = 1, b = 2, c = 3)
 
 # By assigning names() to an existing vector:
+<!-- GVW: "by assigning a character vector to names()" maybe? -->
```
hadley: Can you please move this out of the code block so it doesn't cause the build to fail?

PR #1269, opened 2018-10-14
Fixed pdf output error

slonik-az:
missing {} brackets after '_' subscript in La…TeX formula.

Without this change could not generate pdf output due to latex errors.


hadley:
Thanks!

hadley:
It builds for me without problems.

slonik-az:
Hi Hadley,  
Thanks for merging the PR. Like in your case, gitbook target (html) builds for me fine even without this PR.   But without it building  __`pdf_book`__  results in the following error (from _main.log):
```
This is XeTeX, Version 3.14159265-2.6-0.99999 (TeX Live 2018) (preloaded format=xelatex 2018.4.16)  17 OCT 2018 05:48
...
**_main.tex
...
! Missing { inserted.
<to be read again> 
                   \__um_group_begin: 
l.13367 \(f_\mathbf
                   {x}(\lambda)\) instead of \(f(\lambda, \mathbf{x})\). In R, 
```
which is fixed by this PR (#1269).
My setup is TeXLive-2018 (MacTeX-2018), R-3.5.1, knitr-1.20, pandoc-2.1.1.

Thanks,
--Leo

hadley:
Building `pdf_book` works fine for me too.

PR #1275, opened 2018-10-16
Functions: Exiting a function

lindbrook:


Functions.Rmd
```diff
@@ -951,18 +951,17 @@ Using `...` comes with two downsides:
 
 ## Exiting a function
 
-Most functions exit in one of two ways[^esoterica]: either returning a value, indicating successful completion, or throwing an error, indicating failure. This section describes return values (implicit vs. explicit; visible vs. invisible), briefly discusses errors, and introduces exit handlers, which allow you to run code when a function exits, regardless of how it exits.
+Most functions exit in one of two ways[^esoterica]: it either returns a value, indicating success, or it throws an error, indicating failure. This section describes return values (implicit vs. explicit; visible vs. invisible), briefly discusses errors, and introduces exit handlers, which allow you to run code when a function exits.
```
hadley: Shouldn't the pronoun (it) match "most functions"?

Functions.Rmd
```diff
@@ -1028,13 +1027,13 @@ a <- 2
 (a <- 2)
 ```
 
-And this is what makes it possible to chain assignment:
+This is what makes chain assignment possible:
```
hadley: I can't explain precisely why but this changes the meaning - maybe because it makes "chain assignment" feel like a compound noun, instead of being a verb-noun pair?

PR #1278, opened 2018-10-19
Typo: allows -> always

gavinsimpson:
This PR corrects a trivial typo: `allows` -> `always` as the intention of the sentence seems to be:

> If you do, the method will always return the same class, regardless of the input.


hadley:
Thanks!

PR #1282, opened 2018-10-23
Copy edit S3 chapter

batpigandme:
Minor changes listed below in order of appearance (in case they're ambiguous):

* Missing comma
* Use "a generic" as you do elsewhere
* Missing word
* Subject agreement between two sentences
* Reiterating that it's about validators
* Errant line break
* Conditional comma
* Compound modifier  
* Added comma  
* Missing word  
* Stay in tense


hadley:
Thanks!

PR #1287, opened 2018-10-25
Correction: "women" is not a gender minority

prosoitos:
Change:

"who have made a wonderfully welcoming community for women and other gender minorities"

to:

"who have made a wonderfully welcoming community for women and gender minorities"

At about 50% of the global population, women are not a gender minority :stuck_out_tongue:


hadley:
Thanks!

PR #1300, opened 2018-10-30
Change code to match diagram

prosoitos:


Vectors.Rmd
```diff
@@ -590,7 +590,7 @@ knitr::include_graphics("diagrams/vectors/summary-tree-s3-2.png")
 A data frame is the most common way of storing data in R, and is crucial for effective data analysis. A data frame is a named list of equal-length vectors. It has attributes providing the (column) `names`, `row.names`[^rownames], and a class of "data.frame": 
 
 ```{r}
-df1 <- data.frame(x = 1:2, y = 2:1)
+df1 <- data.frame(x = 1:3, y = letters[1:3])
```
hadley: This will make a factor, which won't match the diagram 😢

prosoitos: Oh! Darn... of course :stuck_out_tongue: That's what working with sensible functions all the time does to someone :wink: One quickly forgets what the less sensible ones do... Sorry about that.

prosoitos: That said, you go over the `stringsAsFactors` argument in the *next* section of the book. I hadn't thought about that when suggesting the change. So if you think that things are better the way they are, even if the diagram does not match the code, of course, feel free to reject the PR.

prosoitos: Whatever you think is more pedagogical

prosoitos: Actually, I have a 3rd suggestion that is, I think, the best: move the diagram into the next section (so after the discussion of `data.frame()` vs `tibble()`. Not only it makes the most sense with what is discussed here, but it actually also matches the pattern that you followed for other sections (for instance your first diagram in the "List" section comes in the 1st sub-section ("Creating lists"). I'll push a new commit with this change.

prosoitos: Done. This is the solution I prefer. Let me know what you think

hadley:
Thanks!

PR #1306, opened 2018-11-05
Match condition-signal bullets to list / chapter order

batpigandme:


Conditions.Rmd
```diff
@@ -75,13 +75,13 @@ There are three conditions that you can signal in code: errors, warnings, and me
 
 * Errors are the most severe; they indicate that there is no way for a function 
   to continue and execution must stop. 
+
+* Warnings fall somewhat in between, and typically indicate that something has
```
hadley: Now needs to say what they fall in between?

PR #1310, opened 2018-11-09
Names-values: Unbinding ... & Answers

lindbrook:


Names-values.Rmd
```diff
@@ -694,43 +694,43 @@ rm(x)
...
 
 ```{r}
 mem_used()
 ```
 
-This number won't agree with the amount of memory reported by your operating system for three reasons:
+This number won't agree with the amount of memory reported by your operating system. There are three reasons:
 
-1. It only includes objects created by R, not the R interpreter itself.
+1. It only includes objects created by R, not the R interpreter.
```
hadley: This makes it sound like it doesn't include objects created by the R interpreter.

PR #1322, opened 2018-11-23
Typo corrections

Tripartio:
Also, for some reason, "## Big ideas" is numbered "15.4 Big ideas", which doesn't seem right, but I don't know how to fix it. Perhaps all subsections here need the tag "{#meta  .unnumbered}"?


hadley:
Thanks - and yes. Do you want to do another PR?

PR #1347, opened 2018-12-12
Change reference link to cppreference

legendre6891:
The C++ community seems to have converged to using cppreference as the "go to" place for online documentation and reference of the standard.  A key advantage is that cppreference is a community wiki -- an open platform as opposed to a closed one -- and generally have higher quality content. Another is that it doesn't have advertisements.

See the discussions here:

- https://stackoverflow.com/questions/6520052/whats-wrong-with-cplusplus-com
- https://meta.stackexchange.com/questions/194788/links-being-changed-to-cppreference-com

While the answers do not completely settle the issue, I find the arguments for preferring cppreference over cplusplus convincing.

### DIFF
--- a/Rcpp.Rmd
+++ b/Rcpp.Rmd
@@ -22,7 +22,7 @@ Typical bottlenecks that C++ can address include:
   implementations of many important data structures, from ordered maps to 
   double-ended queues.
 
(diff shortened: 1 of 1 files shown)


hadley:
Sounds good - thanks!

PR #1354, opened 2018-12-16
Vectors:Introduction

lindbrook:


Vectors.Rmd
```diff
@@ -5,13 +5,13 @@ source("common.R")
 ```
 ## Introduction
 
-This chapter discusses the most important family of data types in base R: the vector types[^node]. You've probably used many (if not all) of the vectors before, but you may not have thought deeply about how they are interrelated. In this chapter, I won't cover individual vectors types in too much depth. Instead, I'll show you how they fit together as a whole. If you need more details, you can find them in R's documentation.
+This chapter discusses the most important family of data types in base R: vectors.[^node] While you've probably already used many (if not all) of the different types of vectors, you may not have thought deeply about how they're interrelated. In this chapter, I won't cover individual vectors types in too much detail, but I will show you how all the types fit together as a whole. If you need more details, you can find them in R's documentation.
```
hadley: The footnote should go before the `.` (at least in my style)

Vectors.Rmd
```diff
@@ -5,13 +5,13 @@ source("common.R")
...
 
-[^node]: Collectively, all other data types are known as the "node" data types, and include things like functions and environments. This is a highly technical term used in only a few places. The place where you're most likely to encounter it is the output of `gc()`: the "N" in `Ncells` stands for nodes, and the "V" in `Vcells` stands for vectors.
+[^node]: Collectively, all the other data types are known as "node" data, which include things like functions and environments. You're most likely to come across this highly technical term when using `gc()`: the "N" in `Ncells` stands for nodes and the "V" in `Vcells` stands for vectors.
```
hadley: I think it would be better to drop "data" than "type" — node type is an important phrase

Vectors.Rmd
lindbrook: OK. That's not even a British-ism, I think.

hadley:
Thanks!

PR #1358, opened 2018-12-18
Comments on translation chapter

gvwilson:
I think I'd need a week to go through the whole metaprogramming suite in detail, working through all the exercises, to really become comfortable with it, but I'm glad it's all in one place.


hadley:
Thanks! The book is due at the publishers at the end of January, but I'm hopeful that I can process all your smaller comments and still have time for some improved diagrams.

PR #1360, opened 2018-12-18
Vectors:Atomic vectors

lindbrook:
Wildcards for is.*() or as.*() functions?

### DIFF
--- a/Vectors.Rmd
+++ b/Vectors.Rmd
@@ -158,7 +158,7 @@ typeof(chr_var)
 \indexc{is.atomic()}
 \indexc{is.numeric()}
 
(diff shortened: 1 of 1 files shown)


lindbrook:
asterik didn't print: "is.*()" etc.

hadley:
`is.*()` I think. In other places I use (e.g.) `geom_` functions, but I think the underscore makes it more clear.

hadley:
Oh I meant functions that already containing an underscore maybe don't need a `*`; i.e. `geom_` functions vs `is.*` functions?

hadley:
Maybe it's better to always include the `*` wildcard?

lindbrook:
I agree. It stands out better and it's _a_ standard.

PR #1369, opened 2018-12-20
Vectors:S3 atomic vectors

lindbrook:


Vectors.Rmd
```diff
@@ -456,7 +456,7 @@ typeof(now_ct)
 attributes(now_ct)
 ```
 
-The `tzone` attribute controls how the date-time is formatted, not the instant of time represented by the vector. Note that the time is not printed if it is midnight.
+The `tzone` attribute controls how the date-time is formatted, not how time is represented by the vector. Note that the time is not printed if it is midnight.
```
hadley: instant is a technical term that I'd prefer to preserve here somehow

Vectors.Rmd
```diff
@@ -361,19 +361,19 @@ str(array(1:3, 3))         # "array" vector
 \index{attributes!S3}
 \index{S3!vectors}
 
-One of the most important attributes is `class`, which defines the S3 object system. Having a class attribute makes an object an __S3 object__, which means that it will behave differently when passed to a __generic__ function. Every S3 object is built on top of a base type, and often stores additional information in other attributes. You'll learn the details of the S3 object system, and how to create your own S3 classes, in Chapter \@ref(s3). 
+One of the most important vector attributes is `class`, which underlies the S3 object system. Having a class attribute turns an object into an __S3 object__, which allows you to do things like use __generic__ functions to determine behaviour. Every S3 object is built on top of a base type, and often stores additional information in other attributes. You'll learn the details of the S3 object system, and how to create your own S3 classes, in Chapter \@ref(s3).
```
```suggestion
 One of the most important vector attributes is `class`, which underlies the S3 object system. Having a class attribute turns an object into an __S3 object__, which means it will behave differently to the underlying vector when passed to generic functions. Every S3 object is built on top of a base type, and often stores additional information in other attributes. You'll learn the details of the S3 object system, and how to create your own S3 classes, in Chapter \@ref(s3). 
```

lindbrook: Something along the lines of : "Having a class attribute turns an object into an __S3 object__, which means it will behave differently from a regular vector when passed to a generic function."

hadley: Yeah, that sounds good.

hadley:
Thanks!

PR #1371, opened 2018-12-21
chunk not exists in purrr 0.2.5 for section 4.3

dongzhuoer:


hadley:
It will exist in the next release of purrr, which is happening soon

PR #1383, opened 2018-12-25
Update Control-flow.Rmd

paleolimbot:
I hope it is not off base to PR here, but I noticed when reading this that the terminology around `for` loops could maybe be more precise, as they don't iterate over indices, but instead iterate over items (elements?) in a vector (which is commonly a vector of indicies). Cheers!


hadley:
Thanks - this feels better.

PR #1388, opened 2018-12-26
Vectors:Data frames and tibbles (intro)

lindbrook:


Vectors.Rmd
```diff
@@ -614,13 +614,13 @@ These data structures are relatively esoteric but they can be useful if you want
 \index{tibbles|see {data frames}}
 \indexc{row.names}
 
-There are two important S3 vectors that are built on top of lists: data frames and tibbles.
+Two of the most important S3 vectors built on top of lists are data frames and tibbles.
```
```suggestion
The two most important S3 vectors built on top of lists are data frames and tibbles. 
```

Vectors.Rmd
```diff
@@ -629,9 +629,9 @@ typeof(df1)
...
+[^rownames]: Row names are one of the most surprisingly complex data structures in R. They've also been a persistent source of performance issues over the years. The most straightforward implementation is a character or integer vector, with one element for each row. But there's also a compact representation for "automatic" row names (consecutive integers), created by `.set_row_names()`. R 3.5 has a special way of deferring integer to character conversion that is specifically designed to speed up `lm()`; see <https://svn.r-project.org/R/branches/ALTREP/ALTREP.html#deferred_string_conversions> for details.
 
-Because each element of the list has the same length, data frames have a rectangular structure, and hence shares properties of both the matrix and the list:
+Because each element of a list has the same length, data frames have a rectangular structure, and hence shares properties of both the matrix and the list:
```
hadley: Maybe "this list"? It's a specific implication about the list underlying data frames, not lists in general.

lindbrook: Right. I forgot to follow up on this point. Above, you go out of your way to emphasize that elements of a list are really just _references_. Thus, the length of each element of a list is really always the same (under the hood). But this creates tension when you write: "A data frame is a named list of equal-length vectors". I'm guessing that the 'data.frame' class imposes an additional constraint that the actual objects being referenced must be of equal length (hence the data frame's rectangular structure). If my understanding is correct, this subtle point, or whatever the true explanation is, is worth a little more exposition (that's why this sentence is a little problematic).

lindbrook: How about: "A data frame is a named list of vectors with the additional constraint that every vector is the same length. This gives data frames their rectangular structure and explains why they share the properties of both matrices and lists."

hadley:
Thanks!

PR #1395, opened 2018-12-28
Vectors: rest of chapter

lindbrook:


Vectors.Rmd
```diff
@@ -754,77 +754,74 @@ rownames(df3)
...
+  use `df[, vars, drop = FALSE]`.
   
 * When you attempt to extract a single column with `df$x` and there is no 
   column `x`, a data frame will instead select any variable that starts with
   `x`. If no variable starts with `x`, `df$x` will return `NULL`. This makes
   it easy to select the wrong variable or to select a variable that doesn't
   exist.
 
-Tibbles tweak these behaviours so that `[` always returns a tibble, and `$` doesn't partial match, and warns if it can't find a variable (this is what makes tibbles surly).
+Tibbles tweak these behaviours: `[` always returns a tibble; `$` doesn't do partial matching and returns a warning if it can't find a variable (this is what makes tibbles surly).
```
hadley: A `,` instead of `;` would feel more natural to me

Vectors.Rmd
```diff
@@ -754,77 +754,74 @@ rownames(df3)
...
+  screen. Additional columns are shown at the bottom.
 
 * Each column is labelled with its type, abbreviated to three or four letters.
 
 * Wide columns are truncated to avoid a single long string occupying an entire
-  row. (This is still a work in progress: it's tricky to get the tradeoff right 
+  row. (This is still a work in progress: it's tricky to get the tradeoff 
   between showing as many columns as possible and showing a single wide 
-  column fully.)
+  column in its entirety just right.)
```
hadley: This doesn't read very well to me (not the original was great either) — I had to re-read to understand what "just right" applied to

Vectors.Rmd
```diff
@@ -887,7 +884,7 @@ data.frame(
 knitr::include_graphics("diagrams/vectors/data-frame-list.png")
 ```
 
-List columns are easier to use with tibbles because you can provide them inside `tibble()`, and they are handled specially when printing:
+List columns are easier to use with tibbles because you can directly include them inside `tibble()`, and because printing them is handled automatically:
```
```suggestion
List columns are easier to use with tibbles because you can directly include them inside `tibble()` and printing is handled automatically:
```

Vectors.Rmd
lindbrook: "Tibbles tweak these behaviours so that a `[` always returns a tibble, and a `$` doesn't do partial matching and warns if it can't find a variable (this is what makes tibbles surly)."

Vectors.Rmd
lindbrook: "Wide columns are truncated to avoid having a single long string occupy an entire
row. (This is still a work in progress: it's a tricky tradeoff 
between showing as many columns as possible and showing columns in their entirety.)"

Vectors.Rmd
lindbrook: "List columns are easier to use with tibbles because they can be directly included inside `tibble()` and they will be tidily printed:"

Vectors.Rmd
hadley: :+1:

Vectors.Rmd
hadley: :+1:

Vectors.Rmd
hadley: printed tidily?

hadley:
Thanks!!

PR #1411, opened 2019-01-05
Subsetting:Intro

lindbrook:


Subsetting.Rmd
```diff
@@ -10,14 +10,14 @@ rownames(mtcars) <- NULL
 
 R's subsetting operators are fast and powerful. Mastering them allows you to succinctly perform complex operations in a way that few other languages can match. Subsetting in R is easy to learn but hard to master because you need to internalise a number of interrelated concepts:
 
-* There are six types of data you can subset.
+* There are six ways to subset atomic vectors.
```
hadley: Here I'm trying to imply that there are six different types of thing that `y` can be in `x[y]`. But I'm using type in a different way to the rest of the book, and this is just the intro, so I think your wording is fine.

hadley:
Thanks!  BTW the book is due at the publishers on Feb 1 — from my perspective it's content complete (although I'm going to try to tackle these [three issues](https://github.com/hadley/adv-r/issues?q=is%3Aissue+is%3Aopen+label%3A%22meta+%3Adragon%3A%22)), but I'll continue to make smaller changes. I don't have much time this month because of rstudio::conf, but I'll try to respond to your changes quickly so we can get it.

PR #1415, opened 2019-01-08
Subsetting: Selecting multiple elements

lindbrook:
recycling: "This is convenient and easy to  understand when one of `x` and `y` is length one, but I recommend avoiding recycling for other lengths because the rules are consistently applied throughout base R."

"inconsistently applied" or "too consistently" (recycling based on multiples of length might not be what you want)


hadley:
Should be inconsistently applied (oops!)

hadley:
Thanks!

In case it ever comes up, I documented the inconsistencies at https://vctrs.r-lib.org/articles/type-size.html#appendix-recycling-in-base-r (but that url is probably not super stable so we shouldn't include in the book)

PR #1419, opened 2019-01-10
Subsetting: Selecting multiple elements ...

lindbrook:


Subsetting.Rmd
```diff
@@ -222,15 +222,15 @@ vals[select]
...
 
-* When subsetting with a single index, they behave like lists, indexing into 
+* When subsetting with a single index, they behave like lists and index 
   the columns, so `df[1:2]` selects the first two columns.
   
 * When subsetting with two indices, they behave like matrices, so
   `df[1:3, ]` selects the first three _rows_ (and all the columns)[^python-dims].
 
-[^python-dims]: If you're coming from Python this is likely to be confusing, as you'd probably expect `df[1:3, 1:2]` to select three columns and two rows. Generally, R seems to "think about" dimensions in the opposite direction to Python.
+[^python-dims]: If you're coming from Python this is likely to be confusing, as you'd probably expect `df[1:3, 1:2]` to select three columns and two rows. Generally, R "thinks" about dimensions in terms of rows and columns while Python do so in terms of columns and rows.
```
hadley:
```suggestion
[^python-dims]: If you're coming from Python this is likely to be confusing, as you'd probably expect `df[1:3, 1:2]` to select three columns and two rows. Generally, R "thinks" about dimensions in terms of rows and columns while Python does so in terms of columns and rows.
```

hadley:
Thanks!

PR #1432, opened 2019-01-24
Control-flow: Loops & Answers

lindbrook:
In Line 38, "`perform_action` is called once for each item in `vector`, rebinding the value of `item` each time.", what do you mean by "rebinding"? incrementing?

### DIFF
--- a/Control-flow.Rmd
+++ b/Control-flow.Rmd
@@ -257,8 +257,8 @@ i
 \indexc{break}
 There are two ways to terminate a for loop early:
 
-* `next` stops the current iteration and advances to the next one.
-* `break` simply exits the entire for loop.
+* `next` exits the current iteration.
+* `break` exits the entire for loop.
 
 ```{r}
 for (i in 1:10) {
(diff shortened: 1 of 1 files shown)


hadley:
I mean that it's like `item <- 1; item <- 2; item <- 3`

Control-flow.Rmd
```diff
@@ -257,8 +257,8 @@ i
 \indexc{break}
 There are two ways to terminate a for loop early:
 
-* `next` terminates the current iteration and advances to the next.
-* `break` exits the entire for loop.
+* `next` stops the current iteration and advances to the next one.
```
hadley: `one` feels weird to me here

lindbrook: How about: `next` stops the current iteration and moves on to the next iteration.

lindbrook: "rebinding" -> "updating"?

hadley: Even iteration feels redundant to me. Updating sounds good

lindbrook: 'next' exits the current loop.
'break' exits the entire for loop.

hadley: next exits the current iteration
break exits the entire for loop
?

hadley:
Thanks!

PR #1435, opened 2019-01-25
metaprogramming: introduction

lindbrook:


Meta.Rmd
```diff
@@ -8,37 +8,37 @@ source("common.R")
...
 
-Closely related to metaprogramming is __non-standard evaluation__, or NSE for short. This term is commonly used to describe the behaviour of R functions, but there are two problems with it. Firstly, NSE is actually a property of an argument (or arguments) of a function, so talking about NSE functions is a little sloppy. Secondly, it's confusing to define something by what it is not (standard), so in this book I'll teach you more precise vocabulary.
+Closely related to metaprogramming is __non-standard evaluation__, NSE for short. This term is commonly used to describe the behaviour of R functions, but there are two problems with this. Firstly, NSE is actually a property of a function's argument (or arguments), so talking about NSE functions is a little sloppy. Secondly, it's confusing to define something by what it's not (standard), so in this book I'll teach you a more precise vocabulary.
```
hadley: it implies the term; this implies the behaviour — and it's the term I have problems with.

Also I like argument coming first because it emphasises the distinction from function.

lindbrook: "Closely related to metaprogramming is __non-standard evaluation__, NSE for short. This term, which is commonly used to describe the behaviour of R functions, is problematic in my opinion. Firstly, NSE is actually a property of the argument (or arguments) of a function, so talking about NSE functions is a little sloppy. Secondly, it's confusing to define something by what it's not (standard), so in this book I'll introduce more precise vocabulary."

hadley: Perfect (although I'd remove "in my opinion")

hadley:
Thanks!

PR #1452, opened 2019-01-30
language: using use quoting -> quoting

hyiltiz:
The language of the sentence could be more better clearer :^)

### DIFF
--- a/Quotation.Rmd
+++ b/Quotation.Rmd
@@ -642,7 +642,7 @@ bquote(-.(xyz) / 2)
   is crucial for functions that evaluate code in the context of a data frame,
   like `subset()` and friends.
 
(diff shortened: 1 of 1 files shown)


Quotation.Rmd
```diff
@@ -642,7 +642,7 @@ bquote(-.(xyz) / 2)
   is crucial for functions that evaluate code in the context of a data frame,
   like `subset()` and friends.
 
-Instead functions that quote an argument use some other technique to allow indirect specification. Rather than using use unquoting all base R approaches selectively turn quoting off, so I call them __non-quoting__ techniques.
+Instead functions that quote an argument use some other technique to allow indirect specification. Rather than unquoting, all base R approaches selectively turn quoting off, so I call them __non-quoting__ techniques.
```
hadley:
```suggestion
Instead functions that quote an argument use some other technique to allow indirect specification. Rather than using unquoting, all base R approaches selectively turn quoting off, so I call them __non-quoting__ techniques.
```

PR #1464, opened 2019-02-11
unifies the order of the atomic vectors types

Nowosad:
Most of the example in this chapter are using the following order: double, integer, logical, character. However, there are two places when the order is different. This PR unifies how the atomic vectors types are explained.

### DIFF
--- a/Vectors.Rmd
+++ b/Vectors.Rmd
@@ -71,7 +71,7 @@ Take this short quiz to determine if you need to read this chapter. If the answe
 \index{numeric vectors} 
 \index{character vectors} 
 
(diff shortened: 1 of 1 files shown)


hadley:
My preferred order is from simplest to most complex, i.e. logical, integer, double, character.

Nowosad:
@hadley I do not want to change your preferred order;) I've been reading the vectors chapter, and found out that you are using the "logical, integer, double, character" order at the beginning, but some diagrams and code examples are shown in the "double, integer, logical, character" order (e.g. see section 3.2.2).

PR #1466, opened 2019-02-13
Fix link to the sloop package

haozhu233:
as it doesn't have a pkgdown site


hadley:
I think I'd prefer to fix this by creating the site: https://github.com/r-lib/sloop/issues/21

PR #1467, opened 2019-02-14
Explicit anchors for part introductions

jennybc:
Makes the URL look like

GOOD, unnumbered:
https://adv-r.hadley.nz/techniques.html

instead of

YUCK:
https://adv-r.hadley.nz/introduction-22.html


hadley:
Nice, thanks!

PR #1468, opened 2019-02-17
Single character typo

csgillespie:
`...lazyeval solves some problems, it create others.`
to
`...lazyeval solves some problems, it creates others.`


hadley:
Thanks! No need to write a summary for very simple changes, I can see them in the diff.

PR #1478, opened 2019-03-06
Fix typo

liao961120:
Fix typo


hadley:
Looks like we fixed elsewhere - thanks for spotting!

juangomezduaso:
There were two typos. The first was fixed elsewhere, but the second was only "half fixed" elsewhere (IMHO).
The duplicated "and" was eliminated but I agree with @liao961120 that an "a" should be inserted there
I´ll make a new pull request for you to apply in case you have overlloked this detail

PR #1482, opened 2019-03-11
Names and values edited

paocorrales:
Unimportant comment: At the beginning of section **2.3.3 List** the editor suggests "Names (i.e. variables) point to values;...". I made the change but I think that the original version sounds better.

### DIFF
--- a/Names-values.Rmd
+++ b/Names-values.Rmd
@@ -289,13 +289,13 @@ knitr::include_graphics("diagrams/name-value/binding-f2.png")
 \indexc{ref()}
 \index{lists}
 
-It's not just names (i.e. variables) that point to values; elements of lists do too. Take this list, which is superficially very similar to the numeric vector above:
+Names (i.e. variables) point to values; elements of lists do too. Consider this list, which is superficially very similar to the numeric vector above:
 
 ```{r list1}
 l1 <- list(1, 2, 3)
 ```
 
-This list is more complex because instead of storing the values itself, it instead stores references to them:
+This list is more complex because instead of storing the values itself, it stores references to them:
 
 ```{r, echo = FALSE, out.width = NULL}
 knitr::include_graphics("diagrams/name-value/list.png")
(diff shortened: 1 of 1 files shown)


Names-values.Rmd
```diff
@@ -289,13 +289,13 @@ knitr::include_graphics("diagrams/name-value/binding-f2.png")
 \indexc{ref()}
 \index{lists}
 
-It's not just names (i.e. variables) that point to values; elements of lists do too. Take this list, which is superficially very similar to the numeric vector above:
+Names (i.e. variables) point to values; elements of lists do too. Consider this list, which is superficially very similar to the numeric vector above:
```
hadley: Yeah, can you please leave the first sentence as is

Names-values.Rmd
```diff
@@ -465,7 +465,7 @@ References also make it challenging to think about the size of individual object
 obj_size(x, y)
 ```
 
-Finally, R 3.5.0 and later have a feature that might lead to surprises: ALTREP, short for __alternative representation__. This allows R to represent certain types of vectors very compactly. The place you are most like to see this is with `:` because instead of storing every single number in the sequence, R just stores the first and last number. This means that every sequence, no matter how large, is the same size:
+Finally, R 3.5.0 and later versions have a feature that might lead to surprises: ALTREP, short for __alternative representation__. This allows R to represent certain types of vectors very compactly. The place you are most like to see this is with `:` because instead of storing every single number in the sequence, R just stores the first and last number. This means that every sequence, no matter how large, is the same size:
```
hadley: "R 3.5.0 and later versions" doesn't sound right to me either, but I can't think of a wording that sounds better.

Names-values.Rmd
```diff
@@ -562,7 +562,7 @@ Together, these two complications make it hard to predict whether or not a copy
 
 \index{loops!performance}
 \index{for loops|see {loops}}
-Let's explore the subtleties with a case study using for loops. For loops have a reputation for being slow in R, but often that slowness is caused by every iteration of the loop creating a copy. Consider the following code. It subtracts the median from each column of a large data frame: 
+Let's explore the subtleties with a case study using `for` loops. `For` loops have a reputation for being slow in R, but often that slowness is caused by every iteration of the loop creating a copy. Consider the following code. It subtracts the median from each column of a large data frame:
```
hadley: I don't think these should be given a code font because I'm using the term "for loop" generally.

Names-values.Rmd
paocorrales: How about this option? "Finally, from R 3.5.0 implemented a feature..." or "Finally, R 3.5.0 and forward have a feature..."

Names-values.Rmd
paocorrales: Done!

Names-values.Rmd
```diff
@@ -738,7 +738,7 @@ This number won't agree with the amount of memory reported by your operating sys
 1. R counts the memory occupied by objects but there may be empty gaps due to 
    deleted objects. This problem is known as memory fragmentation.
 
-## Answers {#names-values-answers}
+## Quiz Answers {#names-values-answers}
```
hadley: I just realised this should be "Quiz answers" because I use sentence case for heading. Can you please do a PR to fix them all?

paocorrales: Yes, we'll do it with the global changes after all the PR are resolved!

PR #1487, opened 2019-03-11
Functions

eliocamp:
* p.110 Should scref, builtin and special be monospace?

* p.110 "They have type builtin or special". Says it's "Unclear; revise: You can specify typeof as builtin or  special". I changed it to "They have typeof..." but I believe that `typeof()` is the function that returns the type property, so I'm not sure if it's correct.  

* p.112. Given a name, like `"mean"` , `match.fun()` lets you find   
Asks to remove quotes around `"mean"`. I interpret that it's on purpose because it's a string so left it as is. 

* p. 121 "you’ll learn about quosures, which reify promises into an R object"   
Reify is marked. The editor didn't know what it meant. After googling it I understand that is correct. I leave it as is. 

* p. 125 I believe "varargs" can be singular. So I left the "a" in "is often called a varargs".  

* p. 134 Asks for quotes to be italicised and more indented. I think this should be done modifying the `quote` LaTeX environment in the preamble.  

* p. 135 I thoroughly support using "their" as gender neutral third person singular but the editor not. I changed it.

### DIFF
--- a/Functions.Rmd
+++ b/Functions.Rmd
@@ -585,7 +585,7 @@ A promise has three components:
     h03(double(x))
     ```
 
-You cannot manipulate promises with R code. Promises are like a quantum state: any attempt to inspect them with R code will force an immediate evaluation, making the promise disappear. Later, in Section \@ref(quosures), you'll learn about quosures, which reify promises into an R object where you can easily inspect the expression and the environment.
+You cannot manipulate promises with R code. Promises are like a quantum state: any attempt to inspect them with R code will force an immediate evaluation, making the promise disappear. Later, in Section \@ref(quosures), you'll learn about quosures, which convert promises into an R object where you can easily inspect the expression and the environment.
 
 
 ### Default arguments
@@ -749,7 +749,7 @@ Because of lazy evaluation, you don't need to worry about unnecessary computatio
 \index{ellipsis|see {...}}
 \index{dot-dot-dot|see {...}}
 
-Functions can have a special argument `...` (pronounced dot-dot-dot). With it, a function can take any number of additional arguments. In other programming languages, this type of argument is often called a _varargs_ (short for variable arguments), and a function that uses it is said to be variadic. 
+Functions can have a special argument `...` (pronounced dot-dot-dot). With it, a function can take any number of additional arguments. In other programming languages, this type of argument is often called _varargs_ (short for variable arguments), and a function that uses it is said to be variadic. 
 
 You can also use `...` to pass those additional arguments on to another function.
 
--- a/Functions.Rmd
+++ b/Functions.Rmd
@@ -150,7 +150,7 @@ sum
 `[`
 ```
 
-They have typeof `builtin` or `special`:
+They have either type `builtin` or type `special`.
 
 ```{r}
 typeof(sum)
@@ -350,7 +350,7 @@ g01()
 
 In this section, you'll learn the formal rules of scoping as well as some of its more subtle details. A deeper understanding of scoping will help you to use more advanced functional programming tools, and eventually, even to write tools that translate R code into other languages.
 
-R uses __lexical scoping__[^dyn-scope]: it looks up the values of names based on how a function is defined, not how it is called. "Lexical" here is not the English adjective relating to words or a vocabulary. It's a technical CS term that tells us that the scoping rules use a parse-time, rather than a run-time structure. 
+R uses __lexical scoping__[^dyn-scope]: it looks up the values of names based on how a function is defined, not how it is called. "Lexical" here is not the English adjective that means relating to words or a vocabulary. It's a technical CS term that tells us that the scoping rules use a parse-time, rather than a run-time structure. 
 
 R's lexical scoping follows four primary rules:
 
@@ -643,7 +643,7 @@ str(h06(10))
 args(sample)
 ```
  
-It looks like both `x` and `size` are required, but in fact if the latter is not supplied, `sample()` uses `missing()` to provide a default. If I were to rewrite sample, I'd use an explicit `NULL` to indicate that `size` is not required but can be supplied:
+It looks like both `x` and `size` are required, but if `size` is not supplied, `sample()` uses `missing()` to provide a default. If I were to rewrite sample, I'd use an explicit `NULL` to indicate that `size` is not required but can be supplied:
 
 ```{r}
 sample <- function(x, size = NULL, replace = FALSE, prob = NULL) {
@@ -749,7 +749,7 @@ Because of lazy evaluation, you don't need to worry about unnecessary computatio
 \index{ellipsis|see {...}}
 \index{dot-dot-dot|see {...}}
 
-Functions can have a special argument `...` (pronounced dot-dot-dot). With it, a function can take any number of additional arguments. In other programming languages, this type of argument is often called a _varargs_ (variable arguments), and a function that uses it is said to be variadic. 
+Functions can have a special argument `...` (pronounced dot-dot-dot). With it, a function can take any number of additional arguments. In other programming languages, this type of argument is often called a _varargs_ (short for variable arguments), and a function that uses it is said to be variadic. 
 
 You can also use `...` to pass those additional arguments on to another function.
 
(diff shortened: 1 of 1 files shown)


hadley:
I think it's mostly too late now (since you're almost done) — but if you do have bigger questions, instead of putting them in the PR comments, it's a bit easier to use inline comments (i.e. `<!-- -->`) in the source Rmd.

Functions.Rmd
```diff
@@ -150,7 +150,7 @@ sum
 `[`
 ```
 
-They have type "builtin" or "special":
+They have typeof `builtin` or `special`:
```
hadley: How about: They have either type `builtin` or type `special`

Functions.Rmd
```diff
@@ -350,12 +350,12 @@ g01()
 
 In this section, you'll learn the formal rules of scoping as well as some of its more subtle details. A deeper understanding of scoping will help you to use more advanced functional programming tools, and eventually, even to write tools that translate R code into other languages.
 
-R uses __lexical scoping__[^dyn-scope]: it looks up the values of names based on how a function is defined, not how it is called. "Lexical" here is not the English adjective "relating to words or a vocabulary". It's a technical CS term that tells us that the scoping rules use a parse-time, rather than a run-time structure. 
+R uses __lexical scoping__[^dyn-scope]: it looks up the values of names based on how a function is defined, not how it is called. "Lexical" here is not the English adjective relating to words or a vocabulary. It's a technical CS term that tells us that the scoping rules use a parse-time, rather than a run-time structure.
```
hadley:
```suggestion
R uses __lexical scoping__[^dyn-scope]: it looks up the values of names based on how a function is defined, not how it is called. "Lexical" here is not the English adjective that means relating to words or a vocabulary. It's a technical CS term that tells us that the scoping rules use a parse-time, rather than a run-time structure. 
```

Functions.Rmd
```diff
@@ -643,7 +643,7 @@ str(h06(10))
 args(sample)
 ```
  
-It looks like both `x` and `size` are required, but in fact if it's not supplied, `sample()` uses `missing()` to provide a default for `size`. If I were to rewrite sample, I'd use an explicit `NULL` to indicate that `size` is not required but can be supplied:
+It looks like both `x` and `size` are required, but in fact if the latter is not supplied, `sample()` uses `missing()` to provide a default. If I were to rewrite sample, I'd use an explicit `NULL` to indicate that `size` is not required but can be supplied:
```
hadley:
```suggestion
It looks like both `x` and `size` are required, but if `size` is not supplied, `sample()` uses `missing()` to provide a default. If I were to rewrite sample, I'd use an explicit `NULL` to indicate that `size` is not required but can be supplied:
```

Functions.Rmd
```diff
@@ -749,7 +749,7 @@ Because of lazy evaluation, you don't need to worry about unnecessary computatio
 \index{ellipsis|see {...}}
 \index{dot-dot-dot|see {...}}
 
-Functions can have a special argument `...` (pronounced dot-dot-dot). With it, a function can take any number of additional arguments. In other programming languages, this type of argument is often called a varargs, and a function that uses it is said to be variadic. 
+Functions can have a special argument `...` (pronounced dot-dot-dot). With it, a function can take any number of additional arguments. In other programming languages, this type of argument is often called a _varargs_ (variable arguments), and a function that uses it is said to be variadic.
```
hadley:
```suggestion
Functions can have a special argument `...` (pronounced dot-dot-dot). With it, a function can take any number of additional arguments. In other programming languages, this type of argument is often called a __varargs__ (short for variable arguments), and a function that uses it is said to be variadic. 
```

Functions.Rmd
```diff
@@ -1106,7 +1106,7 @@ for(i in 1:10) print(i)
 `for`(i, 1:10, print(i))
 ```
 
-Suprisingly, in R, `for` can be called like a regular function! The same is true for basically every operation in R, which means that knowing the function name of a non-prefix function allows you to override its behaviour. For example, if you're ever feeling particularly evil, run the following code while a friend is away from their computer. It will introduce a fun bug: 10% of the time, it will add 1 to any numeric calculation inside the parentheses.
+Suprisingly, in R, `for` can be called like a regular function! The same is true for basically every operation in R, which means that knowing the function name of a non-prefix function allows you to override its behaviour. For example, if you're ever feeling particularly evil, run the following code while a friend is away from his or her computer. It will introduce a fun bug: 10% of the time, it will add 1 to any numeric calculation inside the parentheses.
```
hadley: Per other discussion, please revert to their

Functions.Rmd
```diff
@@ -1176,7 +1176,7 @@ x <- k01(a = 1, 2, 3)
 \index{infix functions} 
 \indexc{\%\%}
 
-Infix functions get their name from the fact the function name comes **in**between its arguments, and hence have two arguments. R comes with a number of built-in infix operators: `:`, `::`, `:::`, `$`, `@`, `^`, `*`, `/`, `+`, `-`, `>`, `>=`, `<`, `<=`, `==`, `!=`, `!`, `&`, `&&`, `|`, `||`, `~`, `<-`, and `<<-`. You can also create your own infix functions that start and end with `%`. Base R uses this pattern to define `%%`, `%*%`, `%/%`, `%in%`, `%o%`, and `%x%`.
+Infix functions get their name from the fact the function name comes between its arguments, and hence have two arguments. R comes with a number of built-in infix operators: `:`, `::`, `:::`, `$`, `@`, `^`, `*`, `/`, `+`, `-`, `>`, `>=`, `<`, `<=`, `==`, `!=`, `!`, `&`, `&&`, `|`, `||`, `~`, `<-`, and `<<-`. You can also create your own infix functions that start and end with `%`. Base R uses this pattern to define `%%`, `%*%`, `%/%`, `%in%`, `%o%`, and `%x%`.
```
hadley: Can you please revert? This suggestion loses the connection to "in"

Functions.Rmd
```diff
@@ -1185,7 +1185,7 @@ Defining your own infix function is simple. You create a two argument function a
 "new " %+% "string"
 ```
 
-The names of infix functions are more flexible than regular R functions: they can contain any sequence of characters except "%". You will need to escape any special characters in the string used to define the function, but not when you call it:
+The names of infix functions are more flexible than regular R functions: they can contain any sequence of characters except the percent symbol (%). You will need to escape any special characters in the string used to define the function, but not when you call it:
```
hadley:
```suggestion
The names of infix functions are more flexible than regular R functions: they can contain any sequence of characters except for `%`. You will need to escape any special characters in the string used to define the function, but not when you call it:
```

Functions.Rmd
```diff
@@ -1230,7 +1230,7 @@ second(x) <- 5L
 x
 ```
 
-I say they "act" like they modify their arguments in place, because, as discussed in [Modify-in-place], they actually create a modified copy. We can see that by using `tracemem()`:
+I say they act like they modify their arguments in place, because, as explained in Section \@ref(modify-in-place) about modify-in-place behaviour, they actually create a modified copy. We can see that by using `tracemem()`:
```
hadley:
```suggestion
I say they act like they modify their arguments in place, because, as explained in Section \@ref(modify-in-place), they actually create a modified copy. We can see that by using `tracemem()`:
```

Functions.Rmd
```diff
@@ -1269,15 +1269,15 @@ names(x)[2] <- "two"
...
+is translated into:
 
 ```{r}
 `*tmp*` <- x
 x <- `names<-`(`*tmp*`, `[<-`(names(`*tmp*`), 2, "two"))
 rm(`*tmp*`)
 ```
 
-(Yes, it really does create a local variable named *tmp*, which is removed afterwards.)
+(Yes, it really does create a local variable named \*tmp\*, which is removed afterwards.)
```
hadley:
```suggestion
(Yes, it really does create a local variable named `*tmp*`, which is removed afterwards.)
```

hadley:
> p.110 Should scref, builtin and special be monospace? 
It seems like srcref should be, since that's the new convention for attributes. I'm not sure about builtin and special

> p.110 "They have type builtin or special". Says it's "Unclear; revise: You can specify typeof as builtin or special". I changed it to "They have typeof..." but I believe that typeof() is the function that returns the type property, so I'm not sure if it's correct.
Suggestion in line.

> p.112. Given a name, like "mean" , match.fun() lets you find
> Asks to remove quotes around "mean". I interpret that it's on purpose because it's a string so left it as is.
Agreed

> p. 121 "you’ll learn about quosures, which reify promises into an R object"
Reify is marked. The editor didn't know what it meant. After googling it I understand that is correct. I leave it as is.
Reify is probably too esoteric. Maybe just "which convert promises into regular R objects" 

> p. 125 I believe "varargs" can be singular. So I left the "a" in "is often called a varargs".
I think I would say "is often called varargs"

> p. 134 Asks for quotes to be italicised and more indented. I think this should be done modifying the quote LaTeX environment in the preamble.
Agreed.

PR #1488, opened 2019-03-11
Environments

eliocamp:
* p. 158 "their environments are called closures, which is why this term is often used interchangeably with function in R’s documentation."    
Asks for "function" to use monospace. Since is not code, I left it as is. 

* p. 168 "which context to return() to when a function completes"    
Asks for "to" to be in monospace. Is not code so I left it as is.


hadley:
Thank you!

PR #1489, opened 2019-03-11
Conditions

eliocamp:
* p. 177 Aks to define web API (similar comment in Subsetting). Is it necessary?


hadley:
I don't think so — thanks!

PR #1493, opened 2019-03-12
Function factories

eliocamp:
* p. 257 Editor asks for an explanation of the Figure

* p. 259 Editor asks to remove quotes around "optimal". I feel that this removes the implication that it's not really "optimal" as stated in the footnote. Changed it to "so-called optimal".


hadley:
* I don't think it needs an explanation as most readers will see how it's generated from the code.
* Agreed, thanks!

PR #1494, opened 2019-03-12
Function operators

eliocamp:
* p.271 Editor asks for "for loop" to be written in monospace font. I left it as is.

### DIFF
--- a/Function-operators.Rmd
+++ b/Function-operators.Rmd
@@ -29,7 +29,7 @@ Function operators are closely related to function factories; indeed they're jus
 
 Function operators are typically paired with functionals. If you're using a for-loop, there's rarely a reason to use a function operator, as it will make your code more complex for little gain.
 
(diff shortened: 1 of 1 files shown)


Function-operators.Rmd
```diff
@@ -29,7 +29,7 @@ Function operators are closely related to function factories; indeed they're jus
 
 Function operators are typically paired with functionals. If you're using a for-loop, there's rarely a reason to use a function operator, as it will make your code more complex for little gain.
 
-If you're familiar with Python, decorators are just another name for function operators.
+If you're familiar with Python, _decorators_ is just another name for function operators.
```
hadley: Why italics here?

eliocamp: It was a correction from the editor. I can revert.

hadley: Yeah, it seems inconsistent because I don't think we use italics in many places

PR #1496, opened 2019-03-12
Base Types

eliocamp:
* p.291 Editor asks to remove hyphen in "object-oriented". The words appear throughout the book with and without hyphens. I don't know if there's a meaningful disctinction between the two. Should we choose a nomenclature and change it throughout?

### DIFF
--- a/base-types.Rmd
+++ b/base-types.Rmd
@@ -8,7 +8,7 @@
 source("common.R")
 ```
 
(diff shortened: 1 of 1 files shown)


hadley:
@batpigandme is my official hyphen consultant. Mara: can we please have a ruling on "object oriented" vs "object-oriented"?

batpigandme:
If treated as a normal compound modifier, it should be hyphenated when applied to a noun (e.g. object-oriented programming; functional programming gets no hyphen, but _functional-programming paradigm_ would, because it's modifying paradigm). So, unless you want to treat it as an exception, I think it should be hyphenated.

hadley:
Thanks @batpigandme! @eliocamp can you please do another PR changing to object-oriented everywhere?

eliocamp:
No problem!

PR #1497, opened 2019-03-12
S3

eliocamp:
* p. 297 Editor asks to change "stats" to "statistics". Left it as is because is the name of the package. 

* p. 298 Changed the names of attributes to monospace and kept the quotes around "factor". 

* p. 310 "It basically creates a vector of method names, paste0("generic", ".", c(class(x), "default")), and then". Should "generic" be quoted? 

* p. 310 Asks to remove quotes around "default". I think that "default class" (without quotes) implies that "default" is an adjective of "class" instead of being the name of the class.

### DIFF
--- a/S3.Rmd
+++ b/S3.Rmd
@@ -233,7 +233,7 @@ The constructor should follow three principles:
 
 * Check the type of the base object and the types of each attribute.
 
-I'll illustrate these ideas by creating constructors for base classes[^base-constructors] that you're already familiar with. To start, lets make a constructor for the simplest S3 class: `Date`. A `Date` is just a double with "Date" as its `class` attribute, and no additional attributes. This makes for a very simple constructor:
+I'll illustrate these ideas by creating constructors for base classes[^base-constructors] that you're already familiar with. To start, lets make a constructor for the simplest S3 class: `Date`. A `Date` is just a double with a single attribute: its `class` is "Date". This makes for a very simple constructor:
 
 [^base-constructors]: Recent versions of R have `.Date()`, `.difftime()`, `.POSIXct()`, and `.POSIXlt()` constructors but they are internal, not well documented, and do not follow the principles that I recommend.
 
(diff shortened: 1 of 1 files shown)


S3.Rmd
```diff
@@ -292,7 +292,7 @@ new_factor(1:5, "a")
 new_factor(0:1, "a")
 ```
 
-Rather than encumbering the constructor with complicated checks, it's better to put them in a separate function. Doing so allows you to cheaply create new objects when you know that the values are correct, and easily re-use the checks in other places.
+Rather than encumbering the constructor, it's better to put complicated checks in a separate function. Doing so allows you to cheaply create new objects when you know that the values are correct, and easily re-use the checks in other places.
```
hadley: I prefer the previous wording. Can you please revert?

S3.Rmd
```diff
@@ -233,7 +233,7 @@ The constructor should follow three principles:
 
 * Check the type of the base object and the types of each attribute.
 
-I'll illustrate these ideas by creating constructors for base classes[^base-constructors] that you're already familiar with. To start, lets make a constructor for the simplest S3 class: `Date`. A `Date` is just a double with a "Date" class attribute, and no additional attributes. This makes for a very simple constructor:
+I'll illustrate these ideas by creating constructors for base classes[^base-constructors] that you're already familiar with. To start, lets make a constructor for the simplest S3 class: `Date`. A `Date` is just a double with "Date" as its `class` attribute, and no additional attributes. This makes for a very simple constructor:
```
hadley: How does this sound?

```suggestion
I'll illustrate these ideas by creating constructors for base classes[^base-constructors] that you're already familiar with. To start, lets make a constructor for the simplest S3 class: `Date`. A `Date` is just a double with a single attribute: its `class` is "Date". This makes for a very simple constructor:
```

S3.Rmd
```diff
@@ -498,7 +498,7 @@ There are two wrinkles to be aware of when you create a new method:
 *   First, you should only ever write a method if you own the generic or the
     class. R will allow you to define a method even if you don't, but it is
     exceedingly bad manners. Instead, work with the author of either the 
-    generic or the class to add the method in their code.
+    generic or the class to add the method in his or her code.
```
hadley: I'm fine with singular they, so can you please revert?

eliocamp: I also like it better. I think there are other instances of "his or her" in other chapters. I can do a global check after everything is merged so it's consistent throughout.

PR #1500, opened 2019-03-13
Adds hyphen to "object oriented"

eliocamp:
Following the advice in #1496, "object oriented something" is now "object-oriented something".


hadley:
Thank you!

PR #1502, opened 2019-03-13
Metaprogramming introduction

eliocamp:
I found that chapter overviews in each introductory part do not share the same style. 
* In introduction to Foundations, is "You use vectors every day in R, so Chapter 3 will dive..."
* In Functional programming is "Chapter 9 shows you how to... "
* In Metaprogramming is "In Big picture, Chapter 17, you’ll get..."
* In Techniques is in paragraphs instead of bullet points.

I don't know if this is intentional but it may be worth revising.


hadley:
Yeah, I should attempt to be consistent. Which style do you prefer?

eliocamp:
I think the "Chapter 9 shows you how to... " style is clear and also consistent with outlines of chapters, where the reader gets "Section 23.2 shows you how to use...". 

We've been compiling a list of global checks that affect multiple chapters [here](https://github.com/eliocamp/adv-r/issues/1). I think they are probably easier to do once every change is merged.

hadley:
Perfect!

PR #1504, opened 2019-03-13
Quasiquotation

eliocamp:
p. 435 Editor asks to remove the las "function" in "Another important class of quoting functions are the base modelling and plotting functions...". I believe is fine as is. 

p. 435 Asks to describe figure. I added a bit, but I'm not sure is enought. 

p. 441 "Now, however, I prefer the approach described next" Where is "next"?


Quotation.Rmd
```diff
@@ -1216,8 +1215,8 @@ The idea of quasiquotation is an old one. It was first developed by the philosop
...
 
-My attempt to resolve these limitations lead to the lazyeval package (2014-2015). Unfortunately, my analysis of the problem was incomplete and while lazyeval solves some problems, it creates others. It was not until I started working with Lionel Henry on the problem that all the pieces finally fell into place and we create the full tidy evaluation framework (2017). Despite the newness of tidy evaluation, I teach it here because it is a rich and powerful theory that, once mastered, makes many hard problems much easier.
+My attempt to resolve these limitations lead to the lazyeval package (2014-2015). Unfortunately, my analysis of the problem was incomplete and while lazyeval solves some problems, it creates others. It was not until I started working with Lionel Henry on the problem that all the pieces finally fell into place and we created the full tidy evaluation framework (2017). Despite the newness of tidy evaluation, I teach it here because it is a rich and powerful theory that, once mastered, makes many hard problems much easier.
```
eliocamp: I'm not sure about the tense in "...while lazyeval solves some problems, it creates others." Shouldn't it be "...lazyeval solved some problems, it created others."?

hadley: I agree a past tense would be better here.

Quotation.Rmd
```diff
@@ -971,6 +971,8 @@ f <- function(..., .dots) {
 
 At the time I discovered it, I found this technique particularly compelling so you can see it used throughout the tidyverse. Now, however, I prefer the approach described next.
 
+<!-- Here is not clear when is "next" -->
```
hadley: I think this should now be "described previously"

Quotation.Rmd
```diff
@@ -272,7 +272,7 @@ There are two other important base quoting functions that we'll cover elsewhere:
 ### Substitution
 \indexc{substitute()}
 
-You'll most often see `substitute()` used to capture unevaluated arguments. However, as well as quoting, `substitute()` also does "substitution": if you give it an expression, rather than a symbol, it will substitute in the values of symbols defined in the current environment. 
+You'll most often see `substitute()` used to capture unevaluated arguments. However, as well as quoting, `substitute()` also handles substitution. If you give it an expression, rather than a symbol, it will substitute in the values of symbols defined in the current environment.
```
hadley:
```suggestion
You'll most often see `substitute()` used to capture unevaluated arguments. However, as well as quoting, `substitute()` also does substitution (as its name suggests!). If you give it an expression, rather than a symbol, it will substitute in the values of symbols defined in the current environment. 
```

Quotation.Rmd
```diff
@@ -290,12 +290,12 @@ substitute(x * y * z, list(x = 10, y = quote(a + b)))
 When quoting (i.e. capturing code), there are two important distinctions: 
 
 * Is it supplied by the developer of the code or the user of the code?
-  I.e. is it fixed (supplied in the body of the function) or varying (supplied
+  Is it fixed (supplied in the body of the function) or varying (supplied
```
hadley:
```suggestion
  In other words, is it fixed (supplied in the body of the function) or varying (supplied
```

Quotation.Rmd
```diff
@@ -290,12 +290,12 @@ substitute(x * y * z, list(x = 10, y = quote(a + b)))
...
   via an argument)?
   
 * Do you want to capture a single expression or multiple expressions?
 
-This leads to a 2 x 2 table of functions for rlang, Table \@ref(tab:quoting-rlang), and base R, Table \@ref(tab:quoting-base).
+This leads to 2 $\times$ 2 tables of functions for rlang, Table \@ref(tab:quoting-rlang), and base R, Table \@ref(tab:quoting-base).
```
hadley:
```suggestion
This leads to a 2 $\times$ 2 table of functions for rlang, Table \@ref(tab:quoting-rlang), and for base R, Table \@ref(tab:quoting-base).
```

Quotation.Rmd
```diff
@@ -728,7 +727,7 @@ There are four basic forms seen in base R:
     \indexc{help()}
 
 \indexc{lm()}
-Another important class of quoting functions are the base modelling and plotting functions, which follow the so-called standard non-standard evaluation rules: <http://developer.r-project.org/nonstandard-eval.pdf>. For example, `lm()` quotes the `weight` and `subset` arguments, and when used with a formula argument, the plotting function quotes the aesthetic arguments (`col`, `cex`, etc):
+Another important class of quoting functions are the base modelling and plotting functions, which follow the so-called standard non-standard evaluation rules: <http://developer.r-project.org/nonstandard-eval.pdf>. For example, `lm()` quotes the `weight` and `subset` arguments, and when used with a formula argument, the plotting function quotes the aesthetic arguments (`col`, `cex`, etc) so that there is no need to use `color = iris$Species`:
```
hadley:
```suggestion
Another important class of quoting functions are the base modelling and plotting functions, which follow the so-called standard non-standard evaluation rules: <http://developer.r-project.org/nonstandard-eval.pdf>. For example, `lm()` quotes the `weight` and `subset` arguments, and when used with a formula argument, the plotting function quotes the aesthetic arguments (`col`, `cex`, etc). Take the following code: we only need `col = Species` rather than `col = iris$Species`.
```

Quotation.Rmd
```diff
@@ -1160,7 +1159,7 @@ Another application of `new_function()` is for functions that work like `graphic
 curve(sin(exp(4 * x)), n = 1000)
 ```
 
-Here `x` is a pronoun: it doesn't represent a single concrete value, but is instead a placeholder that varies over the range of the plot. One way to implement `curve()` is to turn that expression into a function with a single argument, `x`, then call that function:
+In that expresssion, `x` is a pronoun: it doesn't represent a single concrete value, but is instead a placeholder that varies over the range of the plot. One way to implement `curve()` is to turn that expression into a function with a single argument, `x`, then call that function:
```
hadley:
```suggestion
In this code, `x` is a pronoun: it doesn't represent a single concrete value, but is instead a placeholder that varies over the range of the plot. One way to implement `curve()` is to turn that expression into a function with a single argument, `x`, then call that function:
```

hadley:
Thanks!

PR #1506, opened 2019-03-13
Translation

eliocamp:
p. 487 The bold italic text is not showing. The tag functions created by the example fail if they don't have any named arguments because `dots_partition` fails when no argument is named. I added a check in `dots_partition` that fixes the bug.

### DIFF
--- a/Translation.Rmd
+++ b/Translation.Rmd
@@ -172,10 +172,11 @@ We could list all the possible attributes of the `<p>` tag in the function defin
 dots_partition <- function(...) {
   dots <- list2(...)
   
+ if (is.null(names(dots))) {
+  is_named <- rep(FALSE, length(dots))
+} else {
   is_named <- names(dots) != ""
-  if (length(is_named) == 0) {
-    is_named <- rep(FALSE, length(dots))
-  }
+}
   
   list(
     named = dots[is_named],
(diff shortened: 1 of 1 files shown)


Translation.Rmd
```diff
@@ -401,7 +405,7 @@ Before we begin, let's quickly cover how formulas are expressed in LaTeX. The fu
 * Most simple mathematical equations are written in the same way you'd type
   them in R: `x * y`, `z ^ 5`. Subscripts are written using `_` (e.g., `x_1`).
 
-* Special characters start with a `\`: `\pi` = π, `\pm` = ±, and so on.
+* Special characters start with a `\`: `\pi` = $\pi$, `\pm` = $\pm$, and so on.
```
eliocamp: In my pdf, the pi symbol was not printing. Using LaTeX notation should be safer, I think.

hadley: Good call

Translation.Rmd
```diff
@@ -412,7 +416,7 @@ Before we begin, let's quickly cover how formulas are expressed in LaTeX. The fu
   write a fraction you'd use `\frac{a}{b}`. To write a square root, you'd use
   `\sqrt{a}`.
 
-* To group elements together use `{}`: i.e., `x ^ a + b` vs. `x ^ {a + b}`.
+* To group elements together use `{}`: i.e., `x ^ a + b` vsersus `x ^ {a + b}`.
```
hadley:
```suggestion
* To group elements together use `{}`: i.e., `x ^ a + b` versus `x ^ {a + b}`.
```

Translation.Rmd
```diff
@@ -171,8 +171,12 @@ We could list all the possible attributes of the `<p>` tag in the function defin
 ```{r named}
 dots_partition <- function(...) {
   dots <- list2(...)
-
+  
   is_named <- names(dots) != ""
+  if (length(is_named) == 0) {
```
hadley: I think it would be slightly clearer to write:

```R
if (is.null(names(dots))) {
  is_named <- rep(FALSE, length(dots))
} else {
  is_named <- names(dots) != ""
}
```

Translation.Rmd
eliocamp: Is so typical of me to add a typo when editing :rofl:

Translation.Rmd
eliocamp: Yes, it looks better.

PR #1512, opened 2019-03-14
Rewriting R code in C++

eliocamp:


eliocamp:
@hadley in line 644 there's a comment about the user of "sugar". I'm not sure what that means exactly.

Rcpp.Rmd
```diff
@@ -641,6 +641,8 @@ LogicalVector is_naC(NumericVector x) {
 is_naC(c(NA, 5.4, 3.2, NA))
 ```
 
+<!-- Is not clear what "sugar function" means. -->
```
hadley: Oh I think this whole bit can be removed since I removed the large section on sugar functions.

eliocamp: I'll change it to 

``Another alternative is the function `is_na()`, which takes a vector and returns a logical vector.``

hadley: Sorry, I meant remove the entire section that talks about `is_na()`

eliocamp: Oh, ok. I've removed everything after 
````
```{r}
str(missing_sampler())
```
````

PR #1523, opened 2019-03-22
2 typos  in Evaluation.Rmd

juangomezduaso:
The first one already reported and not accepted (I think it was probably overlooked).


hadley:
Thanks!

PR #1525, opened 2019-03-23
Move a sidenote to another section in Evaluation.Rmd

juangomezduaso:
I think that this sidenote is not in the right place. It refers to source2() so I moved it to where I suppose it belongs and makes sense to me, in section 20.2.2 . I appended it to the end of the section, but it could go before the last paragraph as well.


hadley:
Thanks!

PR #1526, opened 2019-03-23
Global changes

eliocamp:
This fixes some global problems. 

* It removes all quotemarks around blockquotes.

* Makes sure there are no "he or she" or "him or her".

* Removes capitalization in "Quiz **A**nswers".

* Rewrites chapter overviews so that they match in style. The OO part didn't have a chapter overview so I wrote one. I don't know exactly *where* it would go, though. 

I need to solve some issues but I need your input:

* Some tables have captions (Tables 19.1 and 19.2) but the rest don't. Should every table (and figure?) have a caption?
* When citing books and other resources sometimes they are around quotes (`...I found "Concepts, Techniques and Models of Computer Programming" [@ctmcp] extremely helpful.`) while others are italicised (`...the terminology of _Extending R_ [@extending-R] and call...`). Should they all be one way or the other? Which one?

### DIFF
--- a/Expressions.Rmd
+++ b/Expressions.Rmd
@@ -380,6 +380,10 @@ The following table summarises the appearance of the different expression subtyp
 
 Both base R and rlang provide functions for testing for each type of input, although the types covered are slightly different. You can easily tell them apart because all the base functions start with `is.` and the rlang functions start with `is_`.
 
+\newpage
+
+<!-- New page so that there's no beak inside the table -->
+
 |                   | base                | rlang                    |
 |-------------------|---------------------|--------------------------|
 | Scalar constant   | —                   | `is_syntactic_literal()` |
--- a/Foundations.Rmd
+++ b/Foundations.Rmd
@@ -4,37 +4,37 @@
 
 To start your journey in mastering R, the following six chapters will help you learn the foundational components of R. I expect that you've already seen many of these pieces before, but you probably have not studied them deeply. To help check your existing knowledge, each chapter starts with a quiz; if you get all the questions right, feel free to skip to the next chapter!
 
-1.  In Chapter \@ref(names-values), you'll learn about an important distinction
+1.  Chapter \@ref(names-values), teaches you about an important distinction
     that you probably haven't thought deeply about: the difference between an 
     object and its name. Improving your mental model here will help you make 
     better predictions about when R copies data and hence which basic 
     operations are cheap and which are expensive.
    
-1.  You use vectors every day in R, so Chapter \@ref(vectors-chap) will dive 
-    into the details, helping you learn how the different types of vector fit 
+1.  Chapter \@ref(vectors-chap) dives into the details of vectors, helping you 
+    learn how the different types of vector fit 
     together. You'll also learn about attributes, which allow you to store 
     arbitrary metadata, and form the basis for two of R's object-oriented 
     programming toolkits.
     
-1.  To write clear, concise, and efficient R code it is important to fully 
-    appreciate the power of subsetting as described in Chapter 
-    \@ref(subsetting). Understanding the fundamental components will allow you 
-    to solve new problems by combining the building blocks in novel ways.
+1.  Chapter \@ref(subsetting) describes how to use subsetting properly to write 
+    clear, concise, and efficient R code. Understanding the fundamental 
+    components will allow you to solve new problems by combining the building 
+    blocks in novel ways.
 
-1.  The tools of control flow allow you to only execute code under certain
-    conditions, or to repeatedly execute code with changing inputs. In Chapter
-    \@ref(control-flow), you'll learn the important `if` and `for` constructs,
-    as well as related tools like `switch()` and `while`.
+1.  Chapter \@ref(control-flow) presents tools of control flow that allow you to
+    only execute code under certain conditions, or to repeatedly execute code 
+    with changing inputs: the important `if` and `for` constructs, as well as 
+    related tools like `switch()` and `while`.
 
-1.  Functions are the most important building block of R code, and in Chapter
-    \@ref(functions), you'll learn exactly how they work, including the 
+1.  Chapter \@ref(functions) deals with functions -- the most important builing 
+    blocks of R code. You'll learn exactly how they work, including the 
     scoping rules, which govern how R looks up values from names. You'll also 
     learn more of the details behind lazy evaluation, and how you can 
     control what happens when you exit a function.
     
-1.  In Chapter \@ref(environments), you'll learn about a data structure that
-    is crucial for understanding how R works, but quite unimportant for data 
-    analysis: the environment. Environments are the data structure that binds 
+1.  Chapter \@ref(environments) describes a data structure that is crucial for 
+    understanding how R works, but quite unimportant for data analysis: the 
+    environment. Environments are the data structure that binds 
     names to values, and they power important tools like package namespaces. 
     Unlike most programming languages, environments in R are "first class" 
     which means that you can manipulate them just like other objects.
--- a/Foundations.Rmd
+++ b/Foundations.Rmd
@@ -4,7 +4,7 @@
 
 To start your journey in mastering R, the following six chapters will help you learn the foundational components of R. I expect that you've already seen many of these pieces before, but you probably have not studied them deeply. To help check your existing knowledge, each chapter starts with a quiz; if you get all the questions right, feel free to skip to the next chapter!
 
-1.  Chapter \@ref(names-values), teaches you about an important distinction
+1.  Chapter \@ref(names-values) teaches you about an important distinction
     that you probably haven't thought deeply about: the difference between an 
     object and its name. Improving your mental model here will help you make 
     better predictions about when R copies data and hence which basic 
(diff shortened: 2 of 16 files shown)


juangomezduaso:
I think line 161 in the proposed  OO chapter overview has a typo ("R4" instead of "R6"), so it should be:
Chapter \@ref(r6) introduces R6, which is a radically different OO system.

eliocamp:
Thanks!

hadley:
I think I only added table captions when I needed to float them because they otherwise caused problems with the flow. Would you mind taking a quick look at the larger tables in the rendered PDF to see if they're causing suboptimal page breaks?

I think the copy editor used italics, so we should be consistent with them.

Foundations.Rmd
```diff
@@ -4,37 +4,37 @@
 
 To start your journey in mastering R, the following six chapters will help you learn the foundational components of R. I expect that you've already seen many of these pieces before, but you probably have not studied them deeply. To help check your existing knowledge, each chapter starts with a quiz; if you get all the questions right, feel free to skip to the next chapter!
 
-1.  In Chapter \@ref(names-values), you'll learn about an important distinction
+1.  Chapter \@ref(names-values), teaches you about an important distinction
```
hadley:
```suggestion
1.  Chapter \@ref(names-values) teaches you about an important distinction
```

Foundations.Rmd
```diff
@@ -4,37 +4,37 @@
...
+    learn how the different types of vector fit 
     together. You'll also learn about attributes, which allow you to store 
     arbitrary metadata, and form the basis for two of R's object-oriented 
     programming toolkits.
     
-1.  To write clear, concise, and efficient R code it is important to fully 
-    appreciate the power of subsetting as described in Chapter 
-    \@ref(subsetting). Understanding the fundamental components will allow you 
-    to solve new problems by combining the building blocks in novel ways.
+1.  Chapter \@ref(subsetting) describes how to use subsetting properly to write
```
hadley:
```suggestion
1.  Chapter \@ref(subsetting) describes how to use subsetting to write 
```

Foundations.Rmd
```diff
@@ -4,37 +4,37 @@
...
+    components will allow you to solve new problems by combining the building 
+    blocks in novel ways.
 
-1.  The tools of control flow allow you to only execute code under certain
-    conditions, or to repeatedly execute code with changing inputs. In Chapter
-    \@ref(control-flow), you'll learn the important `if` and `for` constructs,
-    as well as related tools like `switch()` and `while`.
+1.  Chapter \@ref(control-flow) presents tools of control flow that allow you to
+    only execute code under certain conditions, or to repeatedly execute code 
+    with changing inputs: the important `if` and `for` constructs, as well as
```
hadley:
```suggestion
    with changing inputs. These include important `if` and `for` constructs, as well as 
```

Foundations.Rmd
```diff
@@ -4,37 +4,37 @@
...
-    \@ref(control-flow), you'll learn the important `if` and `for` constructs,
-    as well as related tools like `switch()` and `while`.
+1.  Chapter \@ref(control-flow) presents tools of control flow that allow you to
+    only execute code under certain conditions, or to repeatedly execute code 
+    with changing inputs: the important `if` and `for` constructs, as well as 
+    related tools like `switch()` and `while`.
 
-1.  Functions are the most important building block of R code, and in Chapter
-    \@ref(functions), you'll learn exactly how they work, including the 
+1.  Chapter \@ref(functions) deals with functions -- the most important builing
```
hadley:
```suggestion
1.  Chapter \@ref(functions) deals with functions, the most important building 
```

Meta.Rmd
```diff
@@ -18,28 +18,27 @@ Specifically, this book focuses on tidy evaluation (sometimes called tidy eval f
...
+1.  Chapter \@ref(meta-big-picture) gives a high level description of the whole 
+    metaprogramming story, briefly learning about all major components 
+    and how they fit together to form a cohesive whole.
+
+1.  Chapter \@ref(expressions) shows that that all R code can be described as a 
+    tree. You'll learn how to visualise these trees, how the rules of R's 
+    grammar convert linear sequences of characters into these trees, and how to 
+    use recursive functions to work with code trees.
+
+1.  Chapter \@ref(quasiquotation) presents tools from rlang you can use to
```
hadley:
```suggestion
1.  Chapter \@ref(quasiquotation) presents tools from rlang that you can use to 
```

Meta.Rmd
```diff
@@ -18,28 +18,27 @@ Specifically, this book focuses on tidy evaluation (sometimes called tidy eval f
...
+    tree. You'll learn how to visualise these trees, how the rules of R's 
+    grammar convert linear sequences of characters into these trees, and how to 
+    use recursive functions to work with code trees.
+
+1.  Chapter \@ref(quasiquotation) presents tools from rlang you can use to 
+    capture (quote) unevaluated function arguments. You'll also learn about
+    quasiquotation, which provides a set of techniques to unquote input to make 
+    it possible to easily generate new trees from code fragments.
+
+1.  Chapter \@ref(evaluation) moves to evaluating captured code. Here you'll
```
hadley:
```suggestion
1.  Chapter \@ref(evaluation) moves on to evaluating captured code. Here you'll 
```

Meta.Rmd
```diff
@@ -18,28 +18,27 @@ Specifically, this book focuses on tidy evaluation (sometimes called tidy eval f
...
+    it possible to easily generate new trees from code fragments.
+
+1.  Chapter \@ref(evaluation) moves to evaluating captured code. Here you'll 
+    learn about an important data structure, the  __quosure__, which ensures 
+    correct evaluation by capturing both the code to evaluate, and the 
+    environment in which to evaluate it. This chapter will show you how to put 
+    all the pieces together to understand how NSE works in base R, and how to 
+    write functions that work like `subset()`.
+
+1.  Chapter \@ref(translation) finally combines first-class environments, lexical
```
hadley:
```suggestion
1.  Chapter \@ref(translation) finishes up by combining first-class environments, lexical 
```

OO.Rmd
```diff
@@ -150,3 +150,19 @@ otype(mle_obj)
 ```
 
 Use this function to figure out which chapter to read to understand how to work with an existing object.
+
+
+1.  Chapter \@ref(base-types) first details the basic types upon which the more
```
hadley: I think this should come after the paragraph starting "The goal of this brief introductory chapter"

OO.Rmd
```diff
@@ -150,3 +150,19 @@ otype(mle_obj)
...
+    complex objects are built. 
+    
+1.  Chapter \@ref(s3) introduces S3, which is the most simple and commonly used
```
hadley:
```suggestion
1.  Chapter \@ref(s3) introduces S3, the simplest and most commonly used
```

eliocamp:
I'll check for big figures. You can force float by using the knitr option `fig.env = "figure"`. I can set it the default in the setup chunk, if you'd like.

hadley:
Not figures, tables. I definitely do not want to float most figures.

OO.Rmd
eliocamp: Changed!

eliocamp:
I checked tables and italicised titles of books and stuff.

Expressions.Rmd
```diff
@@ -380,6 +380,10 @@ The following table summarises the appearance of the different expression subtyp
 
 Both base R and rlang provide functions for testing for each type of input, although the types covered are slightly different. You can easily tell them apart because all the base functions start with `is.` and the rlang functions start with `is_`.
 
+\newpage
+
+<!-- New page so that there's no beak inside the table -->
```
hadley: In this case it would be better to float the table (and then reference it in the text)

Foundations.Rmd
```diff
@@ -4,37 +4,37 @@
...
+    components will allow you to solve new problems by combining the building 
+    blocks in novel ways.
 
-1.  The tools of control flow allow you to only execute code under certain
-    conditions, or to repeatedly execute code with changing inputs. In Chapter
-    \@ref(control-flow), you'll learn the important `if` and `for` constructs,
-    as well as related tools like `switch()` and `while`.
+1.  Chapter \@ref(control-flow) presents tools of control flow that allow you to
+    only execute code under certain conditions, or to repeatedly execute code 
+    with changing inputs. These include important `if` and `for` constructs, as well as
```
hadley:
```suggestion
    with changing inputs. These include the important `if` and `for` constructs, as well as 
```

Subsetting.Rmd
```diff
@@ -341,8 +341,8 @@ There are two other subsetting operators: `[[` and `$`. `[[` is used for extract
 
 `[[` is most important when working with lists because subsetting a list with `[` always returns a smaller list. To help make this easier to understand we can use a metaphor:
 
->  "If list `x` is a train carrying objects, then `x[[5]]` is
-> the object in car 5; `x[4:6]` is a train of cars 4-6." 
+>  If list `x` is a train carrying objects, then `x[[5]]` is
```
hadley:
```suggestion
> If list `x` is a train carrying objects, then `x[[5]]` is
```

Expressions.Rmd
eliocamp: I tried, but the problem is that pandoc converts markdown tables to latex using the "longtable" environment, which means that allows tables that span multiple pages and in this case it 
image

From what I've learned, this is backed into pandoc so there's no easy way to change it.

hadley: Ah got it, thanks for the investigation

PR #1531, opened 2019-04-05
Fix typo: in_dir() -> with_dir()

seananderson:
The example above defines `with_dir()`. Looks like the example used to define `in_dir()` and this question needs to be updated.


hadley:
Thanks!

PR #1535, opened 2019-04-29
Update URL for Rcpp example

gweissman:
Apologies for the URL change -- updated static github site to blogdown, will now be stable for the forseeable future. Thanks for the link.


hadley:
Thanks!

PR #1536, opened 2019-04-30
Possible typo in Quiz answers

dralastaircurrie:
The sentence previously read as if 'never coerces strings to factors' related to the print method for tibbles, whereas it should relate to tibbles more generally.


hadley:
Thanks!

PR #1544, opened 2019-05-27
Move inline more prominently into code

MichaelChirico:
A cursory read of the text as is had me scratching my head about where `cfunction` is coming from if not `pryr`.

Closer read found the Prerequisites section, which mentions both `pryr` and `inline`. So it's a bit odd that `library(pryr)` is surfaced but `library(inline)` is not.

So, surfacing it here for future readers like me :)

### DIFF
--- a/C-interface.Rmd
+++ b/C-interface.Rmd
@@ -1,7 +1,3 @@
-```{r, echo = FALSE, message = FALSE}
-library(inline)
-```
-
 # R's C interface {#c-api}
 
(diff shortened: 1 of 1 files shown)


hadley:
You really shouldn't be reading this chapter at all since it's rather out of date (and not linked to from the book site)

MichaelChirico:
still the first result for "r c API" and still has useful examples

hadley:
That gets you to <http://adv-r.had.co.nz/C-interface.html> which is the previous edition of the book, not built live from here.

PR #1546, opened 2019-05-28
minor typo

haleyjeppson:


hadley:
The diff isn't showing the change, which sometimes happens for small changes in a paragraph. Can you please point it out to me?

haleyjeppson:
Yeah, it's difficult to spot since all I did was add an "s". I changed the beginning of the second sentence from "The following table summarise what happens" to "The following table summarises what happens".

hadley:
Good spotting - thanks!

PR #1548, opened 2019-05-30
Possible typo

dralastaircurrie:
Should this be `with_dir()`, to refer to the function defined around line 995 of the chapter? (Or, if it's `in_dir()` in a CRAN package then specify which package?)


hadley:
Thanks!

PR #1551, opened 2019-06-16
Update string pool banana example text

BetweenTwoTests:
Text refers to 1000 while R code refers to 100. Make both consistent to R example


hadley:
Thanks!

PR #1552, opened 2019-06-19
Update Names-values.Rmd

svkucheryavski:
You started tracing on `x`, perhaps the `untracemem()` should be also applied on `x` in this context?


hadley:
Oops, thanks!

svkucheryavski:
Thank you for the great book!

PR #1559, opened 2019-06-29
wrong variable name

martijnvanattekum:
`df` was not defined in the select2 function. I suppose `data` was meant instead. The function does not work when df is not defined in the global env otherwise.


hadley:
Thanks!

PR #1560, opened 2019-06-30
Typo

mustafaascha:
The body referred to `df` but the formal is `data`. 

Though it may be more appropriate to open an issue, it may not merit the attention so I'll mention: I can't tell whether the formals should have a dot before their names (à la line 545). 

It's a pleasure to read these newer chapters, thanks for all of your work!


hadley:
Ooops, thanks!

PR #1562, opened 2019-07-01
Update Functions.Rmd

beanumber:
added reference to Chapter 8; part 1

### DIFF
--- a/Functions.Rmd
+++ b/Functions.Rmd
@@ -209,7 +209,7 @@ In R, you'll often see functions called __closures__. This name reflects the fac
 ### Invoking a function
 \indexc{do.call()}
 
-You normally call a function by placing with arguments wrapped in parentheses: `mean(1:10, na.rm = TRUE)`. But what happens if you have the arguments already in a data structure?
+You normally call a function by placing its name with arguments wrapped in parentheses: `mean(1:10, na.rm = TRUE)`. But what happens if you have the arguments already in a data structure?
 
 ```{r}
 args <- list(1:10, na.rm = TRUE)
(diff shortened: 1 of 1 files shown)


Functions.Rmd
```diff
@@ -209,7 +209,7 @@ In R, you'll often see functions called __closures__. This name reflects the fac
 ### Invoking a function
 \indexc{do.call()}
 
-You normally call a function by placing with arguments wrapped in parentheses: `mean(1:10, na.rm = TRUE)`. But what happens if you have the arguments already in a data structure?
+You normally call a function by placing its name with arguments wrapped in parentheses: `mean(1:10, na.rm = TRUE)`. But what happens if you have the arguments already in a data structure?
```
hadley:
```suggestion
You normally call a function by placing its arguments, wrapped in parentheses, after its name: `mean(1:10, na.rm = TRUE)`. But what happens if you have the arguments already in a data structure?
```

Functions.Rmd
```diff
@@ -948,7 +948,7 @@ j05 <- function() {
 j05()
 ```
 
-An error indicates that something has gone wrong, and forces the user to deal with the problem. Some languages (like C, Go, and Rust) rely on special return values to indicate problems, but in R you should always throw an error. You'll learn more about errors, and how to handle them, in [Conditions].
+An error indicates that something has gone wrong, and forces the user to deal with the problem. Some languages (like C, Go, and Rust) rely on special return values to indicate problems, but in R you should always throw an error. You'll learn more about errors, and how to handle them, in Chapter \@ref(conditions) ([Conditions]).
```
hadley:
```suggestion
An error indicates that something has gone wrong, and forces the user to deal with the problem. Some languages (like C, Go, and Rust) rely on special return values to indicate problems, but in R you should always throw an error. You'll learn more about errors, and how to handle them, in Chapter \@ref(conditions).
```

beanumber:
Looks good to me!

hadley:
Thanks!

PR #1565, opened 2019-07-03
Use `selectMethod` instead of `getMethod`

Marlin-Na:
`selectMethod` should be recommended since `getMethod` does not handle inheritance.


hadley:
Thanks!

PR #1576, opened 2019-09-19
Small error in example, Section 20.3.2

osorensen:
Based on the code in this section, doesn't it make sense to write `eval(get_expr(q1), get_env(q1))` rather than `eval(get_expr(q1), get_env(q2))`? Also corrected a little typo in the same section.


hadley:
Yup, thanks!

PR #1578, opened 2019-10-02
Fix typo in Promises

beanumber:
This works because you defined `x <- 20` on line 473, but it doesn't make any sense in the context of the discussion of promises.


hadley:
Thanks!

PR #1589, opened 2019-12-02
consistent approaches to removing vector names

davechilders:
Currently, the two ways described to remove names from a vector are different:

- `unname()` prints the result, but does not update `x`
- `names(x) <- NULL` updates `x`

A reader may interpret the preceding text "you can remove names from a vector" as `x` being updated. This proposal change explicitly updates `x`.


hadley:
Good call, thanks!

PR #1597, opened 2020-02-15
Back-ref to attributes for pairlists

CorradoLanera:
In section 3.2.1 the first note (16) points to this chapter to explore pairlists when referred to attributes. What about explain that case too, and the reason why they are not lists (or what are the advantages or the limitations)?

I don’t know if/how you would like to expand this chapter; the purposed changes are only to back ref to the chapter which pointed here.

### DIFF
--- a/Expressions.Rmd
+++ b/Expressions.Rmd
@@ -897,10 +897,10 @@ args
 typeof(args)
 ```
 
-2. when working with attributes[^pairlist-attr]
+1. when working with attributes[^pairlist-attr]
 
 [^pairlists-c]: If you're working in C, you'll encounter pairlists more often. For example, call objects are also implemented using pairlists.
-[pairlist-attr]: see \@ref(attributes).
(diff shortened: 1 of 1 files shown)


hadley:
You only need to know that attributes are pairlists when working with them in C, so I'd rather remove the backlink that try and explain here.

CorradoLanera:
Ok to remove the backlink. Anyway, it seems strange to me that I jumped to this section from note 16 in 3.3.1 (Vectors/Attributes/Getting and Settings), where it is written that attributes behave like named lists, but are actually pairlists. Next, in the pairlist section, attributes are not mentioned at all. They are not reported like a "place a user is likely to see pairlist in R", nor they are mentioned in the next C-related note 66.

Alternative purpose: add/modify note 66 someway like: "If you’re working in C, you’ll encounter pairlists more often. For example, call objects and attributes are also implemented using pairlists."

PR #1601, opened 2020-02-27
Update Subsetting.Rmd

CorradoLanera:
A data frame is a column-list (considering previous chapters, we can think cells containing pointers like “the columns” and the objects pointed by the pointers like their content). Moreover, if I think about a “column“ in a data frame I still probably think to (a subset of a) data frame, while if I think about the content of a column in a data frame it is not a (subset of that) data frame for sure!


hadley:
I like it, thanks!

PR #1602, opened 2020-02-27
Update Subsetting.Rmd

CorradoLanera:
Those are true only when subset is performed: a data frame with a single column is (and return) a data frame, as arrays with some dimension with length one don’t drop them.

```
matrix(1:4, nrow = 1)
data.frame(x = 1:4)
```


hadley:
I think the context is already obvious since this is a chapter about subsetting.

PR #1608, opened 2020-04-23
Fix typo: less -> more

mpaulacaldas:
Near the end of the control flow chapter, I think the intention was to say *more* instead of *less*


hadley:
I think the intent is to say less flexible, but still flexible enough.

mpaulacaldas:
Thanks for clarifying! I think this makes it clearer then

PR #1614, opened 2020-05-17
Change output vector initialization

jamespooley:
`simple_map2` works for the xs and ws provided, but using `xs` to set the length of the output vector could cause some unexpected behavior (e.g., 5 `NULL`s in the output vector if making `xs` and `ws` length-3 vectors).


hadley:
Thanks,!

PR #1619, opened 2020-06-09
Python naming convention is `self`, not `this`

aki-nishimura:
Python naming convention for the reference to the object itself is `self`, not `this`.


hadley:
Thanks!

PR #1629, opened 2020-07-30
Emphasize term

salim-b:
I think this contributes to the readability. I had to read the sentence twice before this change (maybe also because EN is not my mother tongue).


hadley:
Thanks!

PR #1632, opened 2020-09-16
Added an example where a warning should definitely be an error

Shelmith-Kariuki:
We are prone to hurriedly converting variables that "look like" numerical data into numeric, without thoroughly investigating to check whether there are some "illegal" characters that exist e.g. commas, "+", hyphens, etc. These leads to NAs which should have been avoided in the first place. We are often prone to ignoring warnings, but this is an "error" that should never be ignored. 
image

### DIFF
--- a/Conditions.Rmd
+++ b/Conditions.Rmd
@@ -182,6 +182,8 @@ formals(1)
 file.remove("this-file-doesn't-exist")
 
 lag(1:3, k = 1.5)
+
+as.numeric (c("18", "30", "50+", "345,678"))
 ```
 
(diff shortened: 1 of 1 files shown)


Conditions.Rmd
```diff
@@ -182,6 +182,8 @@ formals(1)
 file.remove("this-file-doesn't-exist")
 
 lag(1:3, k = 1.5)
+
+as.numeric (c("18", "30", "50+", "345,678"))
```
hadley, line 186:
```suggestion
as.numeric(c("18", "30", "50+", "345,678"))
```

hadley:
Good idea, thanks!

PR #1633, opened 2020-10-22
Is "unique names" enough?

MarceloRTonon:
In a list you can have objects that are not named, but in environment you cannot as you say latter in the Quiz Answer as "every object in an environment must have a name". As a non native speaker I may be wrong, but saying "Every name must be unique", only imply that you cannot have two objects with the same name, not that you cannot have a object without names. 

So, in my understanding, saying in section 7.2 that  "Every name must be unique" cannot be easily translated to "every object in an environment must have a name". Therefore I think that adding "and all of its objects must  be binded to a name" serves it well.

This is undirectly explained through basics, but I think that changing that sentence prepares more the reader to the following of the chapter.

Thanks for your time!

### DIFF
--- a/Environments.Rmd
+++ b/Environments.Rmd
@@ -61,7 +61,7 @@ The `env_` functions in rlang are designed to work with the pipe: all take an en
 
 Generally, an environment is similar to a named list, with four important exceptions:
 
-*   Every name must be unique.
+*   Every name must be unique and all of its objects be binded to a name.
 
 *   The names in an environment are not ordered.
 


Environments.Rmd
```diff
@@ -896,7 +898,7 @@ As well as powering scoping, environments are also useful data structures in the
 
 ## Quiz answers {#env-answers}
 
-1.  There are four ways: every object in an environment must have a name;
```
MarceloRTonon, line 899: Even if my alterations are not implemented, this must be changed to be consistent with #734

MarceloRTonon:
Okay, I just read the issue #734 and I understand that there is a delicate situation here:

- Saying that every object in an Environment must have an unique name is **conceptualy wrong**. 

- It is also not recommendable to leave as it is today. It takes some reading to undo the confusion that this makes.

To say that "Every name must be unique", make it seems that, in this matter, the difference between lists and environments is just that two elements of a list can have the same name, but not in an environment. Yes, this is truth, but is an incomplete truth.


As it is written today, it makes you think that the difference is that `l` can have 2 elements named "a", while `e` cannot.

``` r
l <- list("a" =1, "a" =2, "b" = 3)
  
e <- rlang::env("a" = 1, "a" = 2, "b" = 3)
  
l
#> $a
#> [1] 1
#> 
#> $a
#> [1] 2
#> 
#> $b
#> [1] 3
 
rlang::env_print(e)
#> <environment: 0000000013A104A8>
#> parent: <environment: global>
#> bindings:
#>  * a: <dbl>
#>  * b: <dbl>

```
<sup>Created on 2020-10-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This does not makes it simplier to the reader understand that an environment is the set of names that binds objects with names. It only makes it seem that an environment is list that does not allow repeated names (and with the other elements, that it does not order it). Therefore I suggest that you add a fifth

PR #1639, opened 2020-11-14
what is ".y."

xiaochi-liu:
Actually, I don't quite understand this part: .x and .y. for two argument functions, and ..1, ..2, ..3, etc, for functions with an arbitrary number of arguments.

### DIFF
--- a/Functionals.Rmd
+++ b/Functionals.Rmd
@@ -192,7 +192,7 @@ This works because all purrr functions translate formulas, created by `~` (prono
 as_mapper(~ length(unique(.x)))
 ```
 
-The function arguments look a little quirky but allow you to refer to `.` for one argument functions, `.x` and `.y.` for two argument functions, and `..1`, `..2`, `..3`, etc, for functions with an arbitrary number of arguments. `.` remains for backward compatibility but I don't recommend using it because it's easily confused with the `.` used by magrittr's pipe.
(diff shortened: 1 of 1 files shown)


hadley:
Thanks, that was just a typo.

xiaochi-liu:
Hello Hadley,

May I ask a question about this sentence in the section of "9.2.2 Anonymous functions and shortcuts":

>  The function arguments look a little quirky but allow you to refer to `.` for one argument functions, `.x` and `.y` for two argument functions, and `..1`, `..2`, `..3`, etc, for functions with an arbitrary number of arguments.

Does it mean when I define an inline anonymous function in `map()`, if the anonymous function has: 

- 1 argument: I can use `.` or `.x`;
- 2 arguments: I can use `.x` and `.y`;
- 3 arguments and above: perhaps instead of the inline anonymous function, I should use an independent function definition?

If so, what's the meaning of `..1`, `..2`, `..3`, etc, for functions with an arbitrary number of arguments? Could you please provide an example?

Many thanks for your guidance!

hadley:
I think it's easiest to see by looking at what `as_function()` does:

``` r
rlang::as_function(~ 1)
#> <lambda>
#> function (..., .x = ..1, .y = ..2, . = ..1) 
#> 1
#> attr(,"class")
#> [1] "rlang_lambda_function" "function"
```

<sup>Created on 2020-11-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

Then `..1`, `..2` etc are just the standard (if esoteric) R syntax for accessing specific elements of `...`.

xiaochi-liu:
Ok, get it. Thank you very much, Hadley.

PR #1640, opened 2020-11-14
Variable name change: "intercepts" to "slopes"

xiaochi-liu:
I think this variable is better to be called "slopes" since the function is extracting the coefficient of "wt".


hadley:
Good catch — thanks!

PR #1642, opened 2020-11-19
Tighten `chatty()` example

dkahle:
I may be missing a deeper point here, but it seems simpler for `chatty()` to use `message()` instead of `cat()` and call it before the processing begins?

### DIFF
--- a/Function-operators.Rmd
+++ b/Function-operators.Rmd
@@ -14,9 +14,8 @@ chatty <- function(f) {
   force(f)
   
   function(x, ...) {
-    res <- f(x, ...)
-    cat("Processing ", x, "\n", sep = "")
-    res
+    message("Processing ", x)
+    f(x, ...)
   }
 }
 f <- function(x) x ^ 2


hadley:
Hmmm, that is weird. Is `chatty()` used later in the chapter?

dkahle:
No, `grep`'s only showing three instances of `chatty`, two in that first code chunk, and one in the preceding paragraph:
```bash
~/Dropbox/dev/adv-r/adv-r(master*) » grep 'chatty' Function-operators.Rmd
In this chapter, you'll learn about function operators. A __function operator__ is a function that takes one (or more) functions as input and returns a function as output. The following code shows a simple function operator, `chatty()`. It wraps a function, making a new function that prints out its first argument. You might create a function like this because it gives you a window to see how functionals, like `map_int()`, work.
chatty <- function(f) {
purrr::map_dbl(s, chatty(f))
```

hadley:
Thanks!

PR #1659, opened 2020-12-22
Typo fix in 14.3.1

JavOrraca:
In the paragraph prior to 14.3.2 Active fields, this edit would replace "[what] it's [ok]" with "[what] is [ok]" in the following sentence: "The distinction between public and private fields is important when you create complex networks of classes, and you want to make it as clear as possible what is ok for others to access."


hadley:
Thanks!

PR #1660, opened 2020-12-28
Update Vectors.Rmd to avoid object names overwriting base functions

aakosm:
Changed the example object names in the first chunk in https://adv-r.hadley.nz/vectors-chap.html#attr-dims, in order to avoid giving a function name to an object `c <- ...` I renamed all the examples in the chunk from a, b, c to x, y, z.


hadley:
Thanks!

PR #1664, opened 2021-01-16
Update Names-values.Rmd

hukai916:
Tested with R 4.0.2, no copy was made when using list.


the-mad-statter:
This is due to the NAMED -> reference counting change in R v4.0.0. The 15 copies example should also be changed. See my answer on [StackOverflow](https://stackoverflow.com/questions/61844416/example-in-advanced-r-of-modifying-a-list).

hadley:
Thanks!

PR #1667, opened 2021-01-29
Fix typo in Factories exercise: model --> formula

milanmlft:
The `boot_model()` function does not have a `model` argument. It does have `formula`.

https://github.com/hadley/adv-r/blob/1ab83672ad077400b842938804a3773f4279115c/Function-factories.Rmd#L554-L563


hadley:
Thanks!

PR #1668, opened 2021-02-10
Removed unnecessary class 'sidebar'

HannesOberreiter:
In the main content there is part of the text wrapped inside a `sidebar` class, which sets max width, if > 1200px. I don't know if this class is important for the book export as PDF but at least for the browser version it does make no difference if we remove it.

If it is important for the PDF export, it would make sense to write a css exception for `sidebar` inside `content` class.

sidebar class content

Cheers
Hannes


MarceloRTonon:
Nice, this also happens in 20.2.2

https://github.com/MarceloRTonon/adv-r/commit/fd72b80609a402017a42209a7bb5f02374f17117

PR #1669, opened 2021-02-13
Refine void_tag() description

jameslairdsmith:
Slight change of language to say that functions generated by `void_tag()` error when given "child tags", rather than "unnamed tags" and that those child tags are captured by "unnamed dots".


hadley:
Thanks!

PR #1672, opened 2021-03-02
Replace paste0 with paste in Control-flow.Rmd

danielbasso:
Hello! I think `paste0()` is being misused in the birthday example. `paste0` is literally `paste()` with `sep=""`; why use it and then manually put blanket spaces at the strings?  I know it is a silly "correction", but this section of the book is all about the little details, tricks and optimizations that can make your code cleaner and more efficient. I think this example is going against that philosophy; kinda like using `x[which(y)]` instead of `x[y],` as you describe in chapter 4. (Btw I think that `paste0()` is the ugliest function name in all R and prefer to use `paste(sep="")`, just to make the code nicer :) ).

Have all a nice day!

### DIFF
--- a/Control-flow.Rmd
+++ b/Control-flow.Rmd
@@ -77,9 +77,9 @@ When you use the single argument form without an else statement, `if` invisibly 
 
 ```{r}
 greet <- function(name, birthday = FALSE) {
-  paste0(
-    "Hi ", name,
-    if (birthday) " and HAPPY BIRTHDAY"
+  paste(
+    "Hi", name,
+    if (birthday) "and HAPPY BIRTHDAY"
   )
 }
 greet("Maria", FALSE)


danielbasso:
Ok, I realized now that `paste("Hi", "Maria", NULL)` returns `"Hi Maria "`, with a blank space at the end. Which for me creates a more interesting discussion, as the book states that:

> Since functions like `c()` and `paste()` drop `NULL` inputs, this allows for a compact expression of certain idioms

I mean, NULL in this case is interacting with `paste()`, resulting in a blank space.

For example, `paste("Hello", NULL, NULL, NULL, "Maria", sep = "-")` returns `"Hello----Maria"`; this is interesting as the above quote kinda implies that `NULL`s will be ignored in this scenario.

hadley:
I don't think you should ever use `paste()` — it's always better to be explicit about where the spaces are.

PR #1674, opened 2021-03-24
Taking out sidebar

MarceloRTonon:
In the part about Expression Vectors of section 20.2.2, the formatting restricts lines to less than half the page, being unfriendly to the readers.

In the PC it displays as bellow (zoom 50% less):

advR_bad_format


hadley:
This needs to be replaced with something else, not just removed.

PR #1676, opened 2021-04-21
Update Functions.Rmd - removed sidebar class

HannesOberreiter:
Same style problem as https://github.com/hadley/adv-r/pull/1668 and https://github.com/hadley/adv-r/pull/1674. 

Cheers
Hannes


hadley:
I doubt removing the sidebar class is the right thing to do. It needs to be replaced with an Rmd note class or similar.

PR #1687, opened 2021-05-27
Update OO.Rmd

HannesOberreiter:
Added comment for a different terminology for functional OOP or very similar concept which did help me to understand it more.


hadley:
As I understand it, multiple dispatch refers to something a bit different

PR #1693, opened 2021-07-30
Update Rcpp.Rmd - Overwrite wrong auto-links to R Functions

HannesOberreiter:
The `find()` and `count()` function was linking to the R function definition, it should in this case link to the corresponding C++ unordered set methods.


hadley:
Thank you!

PR #1694, opened 2021-08-01
stringsAsFactors defaults to FALSE as of R 4.0.0

debruine:
Under #df-create, the first example doesn't show the character column as a factor, presumably because this was written under R < 4.0.0 but now run under R >= 4.0.0.


hadley:
Thanks!

PR #1698, opened 2021-08-17
boostrap_summary() -> bootstrap_summary()

schoonees:
Fixed typo in function name of one example


hadley:
I can never spell that right. Thanks!

PR #1699, opened 2021-08-19
Update Perf-measure.Rmd - profvis torture flag

HannesOberreiter:
For me the exercise was not working as the GC was called too often? It is maybe only because my computer is too slow? Therefore I changed the torture to a higher integer than TRUE (1).

The torture flag takes an integer `profvis::profvis(torture = step)`  as in the background it uses `base::gctorture2` 

https://github.com/rstudio/profvis/blob/master/R/profvis.R
https://stat.ethz.ch/R-manual/R-devel/library/base/html/gctorture.html

Result from the exercise, with `torture=10`.

profvis exersice


hadley:
Thanks!

PR #1709, opened 2021-11-18
Typo

mathijsdeen:
The assumingly correct argument name `bootstrap_summary` (notice `boot`, not `boo`) occurs inline just below this code chunk (line 296).


mathijsdeen:
This apparently corresponds with the earlier [#1698](https://github.com/hadley/adv-r/pull/1698) so I will close the pull request.

PR #1712, opened 2021-12-01
use expect_identical over expect_equal

MichaelChirico:


hadley:
Equal is fine here, since we don't care about floating point?

PR #1714, opened 2021-12-09
Update Functionals.Rmd

PursuitOfDataScience:
It should be "draw." No reason to use past tense.


hadley:
The present tense doesn't sound right to my ear.

PR #1718, opened 2021-12-15
Corrected small typo

osorensen:


osorensen:
I noticed that this is fixed by PR #1684.

PR #1725, opened 2022-04-27
Informing about the operator `:::`

prdm0:
I think it would be good, in this section, to inform the role of the `:::` operator. The operator is used more explicitly in Section 10.3.3. Informing in the Section where the S3 object-oriented system is discussed would, in my opinion, be a good opportunity to make clear to the reader the usefulness of the `:::` operator.

Best regards.


hadley:
The closest base R equivalent is `getS3method()`, but I don't think it's that useful to mention here.

PR #1727, opened 2022-05-13
Expression vectors no sidebar

dragosmg:
Having the **Expression vectors** section in a _sidebar_ makes it a bit more difficult to read.


dragosmg:
This fails at:

https://github.com/hadley/adv-r/blob/dc49c3872c3530ac08716fd4f4c235b01266a4ce/Control-flow.Rmd#L106

given that the latest version of R (4.2) errors now if the condition has length > 1.

hadley:
This needs to be replaced with something else, not just removed.

PR #1729, opened 2022-06-23
Typo

vedhav:
The foundations talk about `seven` chapters but it is mentioned as `six` in the introduction paragraph.


vedhav:
Didn't realise there was already a PR for this #1724

PR #1749, opened 2022-11-12
Edit in Subsetting chapter

mmhamdy:
A small edit to note that `[[` can do partial matching too, but not by default.


hadley:
Thanks!

PR #1752, opened 2023-01-09
Update status of Martin Morgan

llrs:
The [contributors page](https://www.r-project.org/contributors.html) says explains that Martin is no longer an R core member:

>  Martin Morgan up to June 2021.

This PR is just to reflect that


hadley:
Thanks!

PR #1754, opened 2023-01-17
Update Function-operators.Rmd

taroyabuki:
It says that the first two values are defined by convention, but thanks to this definition we have the beautiful theorem GCD(F(m), F(n)) = F(GCD(m, n)).


hadley:
Thanks!

PR #1759, opened 2023-02-08
Conditions of length greater than 1 result in an error since R 4.2.0

heike:
Any conditions in if statements that have a length of > 1 are considered errors by default since R 4.2.0. Thanks, @HenrikBengtsson! 
https://github.com/HenrikBengtsson/Wishlist-for-R/issues/38

The value of _R_CHECK_LENGTH_1_CONDITION_ will no longer be checked.  This means that regardless of whether the condition is set or not, the command if (c(TRUE, FALSE)) 1
will result in an error.


hadley:
And this one 😞  I guess I should've looked at the PRs before fixing the build problem myself!

PR #1760, opened 2023-02-10
wrap() has been removed from knitr 1.40 in favor of sew()

heike:
From NEWS.md: https://github.com/yihui/knitr/blob/master/NEWS.md
> The internal function knitr:::wrap() has been removed from this package. If you rely on this function, you will have to use the exported function knitr::sew() instead.

And I sign off on the IP :)


hadley:
Oops, just did this myself.

PR #1765, opened 2023-03-19
Update README.md

SimonWaldherr:
the workflow was renamed by [this commit](https://github.com/hadley/adv-r/commit/9081bcd221355ce358a6a1031f04f9c15daf36a3), so the badge has to be adjusted as well.


hadley:
Thanks!

